PR #4, opened 2019-11-21
simplify fetchPypi, remove need for overrides

FRidh:
We already have the name of the file that we need, so let's fetch it
directly instead of breaking it apart first, passing it to the Nixpkgs
fetchPypi, which would then put it together again.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,6 +14,18 @@ let
     else default
   );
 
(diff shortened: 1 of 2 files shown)


adisbladis:
There was a minor issue with how the URLs were constructed which you likely didn't notice because it was a fixed-output derivation.

PR #6, opened 2019-11-21
poetry: make it manylinux compatible

FRidh:
By having the python used for poetry manylinux compatible, it becomes
easier to create lock files in case of extension modules.

The Nixpkgs Python interpreters need to improve on this front, but until
then let's just override them and force compatibility.

https://github.com/NixOS/nixpkgs/issues/71935


adisbladis:
Closing in favour of https://github.com/nix-community/poetry2nix/pull/18

PR #8, opened 2019-11-27
Fixes pypi fetching

gilligan:
- The first letter path segment in the url needs to be lowercased

  https://files.pythonhosted.org/packages/source/S/sqlalchemy/SQLAlchemy-1.3.11.tar.gz
  vs
  https://files.pythonhosted.org/packages/source/s/sqlalchemy/SQLAlchemy-1.3.11.tar.gz

- Rename fetchPypi to fetchFromPypi just to make it clear that we are
  not using fetchPypi from nixpkgs

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,15 +14,19 @@ let
     else default
   );
 
-  # Fetch the artifacts from the PyPI index.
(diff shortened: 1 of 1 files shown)


adisbladis:
Thanks!

PR #10, opened 2019-11-28
Add more tests

gilligan:
Add more tests
    
This adds the following new tests:
    
 - trivial:           minimaltest case
 - override-support:  verify overriding
 - common-pkgs-1/2:   verify commonly downloaded pypi pkgs can be built
    
Also adds some overrides as needed by some of the tested packages.I top downloaded pypi packages have been extracted from https://hugovk.github.io/top-pypi-packages/ - while this doesn't proof correctness checking that the most common packages can be parsed/built seems like a good indicator at least.


adisbladis:
Made a super minor fixup and removed trailing white space, pushed to master in https://github.com/adisbladis/poetry2nix/commit/1bfd3d192bd62c6308491f4cefa8bd73dd4fbdf7.

Thanks!

PR #15, opened 2019-12-11
nixpkgs-fmt: format all the nix code

zimbatm:


gilligan:
I really wish i had been given the opportunity to give feedback on this. I would have definitely tried to weigh in with a 👎

It really makes various places of the code just _so_ much harder to read :( Plus it kinda screws me over with my current soon to be published PR for manylinux wheel support. Much more effort to rebase now.

I don’t know to what degree the output of the formatting depends on some line endings and what not but the current state is imho worse than before.

gilligan:
PS: @zimbatm There is a code change in here. the `intreehooks` stuff. Maybe that slipped into this PR by accident? Would have been very nice to put this into a separate since this doesn't have anything to do with formatting

zimbatm:
One trick is to run `nixpkgs-fmt` on your branch, that should make it easier to rebase it.

For the rest, please try again with some constructive feedback. Frankly it's quite toxic.

gilligan:
@zimbatm that is the way I ended up approaching it yes - thanks for the suggestion.

> For the rest, please try again with some constructive feedback. Frankly it's quite toxic.

I certainly didn't mean to offend you. Please accept my apology if I did though none the less. That being said, could you elaborate on what you perceived as toxic? 

The main points I _attempted_ to make were

1. I would have preferred to not have the nixpkgs-fmt formatting merged. It seems to me that it makes parts of the code _less_ readable. Plus it appears to be one time change. The tool hasn't been added anywhere and no additional checks in the ci I think? At least that hasn't been mentioned anywhere
2. The main job of this PR appears to be to apply the nixpkgs-fmt formatting. However there are also changes that are completely unrelated. I was wondering if they were added intentionally. There isn't any description or anything so in my opinion it would have been better to split these changes from the PR.

Thanks.

PR #17, opened 2019-12-12
CI checks

zimbatm:
Add more CI checks as promised

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,7 @@
 test:
   script:
     - nix-build --no-out-link --show-trace tests
+
+nixpkgs-fmt:
+  script:
+    - ./check-fmt
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,13 @@
-test:
+test-19.09:
+  variables:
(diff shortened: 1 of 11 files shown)


pkgs/intreehooks/default.nix
```diff
@@ -0,0 +1,28 @@
+{ lib
+, buildPythonPackage
+, fetchPypi
+, pytoml
+, pytest
+}:
+
+buildPythonPackage rec {
+  pname = "intreehooks";
```
adisbladis: No need to add `intreehooks` to the overlay, I've backported it to 19.09.

zimbatm:
@adisbladis fixed

pkgs/intreehooks/default.nix
zimbatm: maybe it's worth keeping it?

>  attribute 'intreehooks' missing, at /builds/nix-community/poetry2nix/default.nix:209:34
(from CI)

zimbatm:
@adisbladis re-added intreehooks to fix the CI

pkgs/intreehooks/default.nix
zimbatm: re-added the commit

PR #18, opened 2019-12-14
Add manylinux support

gilligan:
**Description**

This PR adds manylinux support to poetry2nix. It depends on a nixpkgs revision that has the following changes:

- https://github.com/NixOS/nixpkgs/pull/73866 : Adds manylinux packages
- https://github.com/NixOS/nixpkgs/pull/75596 : Use lib outputs of dependencies
- https://github.com/NixOS/nixpkgs/pull/75763 : Removes `_manylinux.py` files from python interpreters

**Additional Info**

- **selectWheel**: The added `pep425.nix` provides a `selectWheel` function which picks a preferred `.whl` based on python version, arch and some heuristics. However `selectWheel` is only ever used if no source distribution is available.

- **mkPoetryEnv**: Provides a python environment with the packages specified by poetry. This is useful in scenarios when you don't want to actually build a python application per se but instead want to provide an python environment to work with. 

- **Refactoring** The PR also contains quite some refactoring which aims to make the code base easier to use with. On several occasions code was extracted, let bindings reduced or some functions could be replaced with builtin or lib functions.

---

Some additional work was required to get this over the finish line... 

1. The first thing that needed to be extended was the parsing of file names according to PEP425. Files using `py2` or `py3` and especially `py2.py3` would not be picked up. I fixed all of that and added proper tests for those scenarios.
1. Furthermore we added `mkPoetryEnv` where all packages specified in the poetry lock file are readily available. If you use `mkPoetryPython` without actually forcing/referencing to some *poetry* dependency you will get a vanilla python package instead.
1. I noticed that all over the project there was code passing `poetryLock` to `mkPoetryApplication` which doesn't even exist. However the combination of a default argument and `...` for additional arguments meant this remained unnoticed. I removed the default arguments and also removed `src` from `mkPoetryPython` and `mkPoetryEnv` which both don't need it.
1. I also updated the README.. (I turned it into markdown..hope that's ok..)


gilligan:
@zimbatm @adibladis care to have a quick look already? Especially the refactoring 

@andir and I put some effort into it and there is a small part left still. Obviously this will get harder to keep up to date once more stuff gets merged. I think the refactoring should be quite beneficial and make the code easier to work with..

default.nix
```diff
@@ -1,10 +1,18 @@
 { pkgs ? import <nixpkgs> {}
 , lib ? pkgs.lib
 , poetry ? null
+, poetryLib ? import ./lib.nix { inherit lib; }
```
zimbatm: what's the advantage of making the `lib` and `poetryLib` parameters overridable?

lib.nix
```diff
@@ -0,0 +1,37 @@
...
+    operators = {
+      "||" = cond1: cond2: cond1 || cond2;
+      "," = cond1: cond2: cond1 && cond2; # , means &&
+    };
+    # split string at "," and "||"
+    tokens = builtins.filter (x: x != "") (builtins.split "(,|\\|\\|)" pythonVersions);
+  in
+    (
+      builtins.foldl' (
+        acc: v: let
```
zimbatm: try moving the `acc: v:` function into the parent `let` binding, it's going to make it more readable

lib.nix
```diff
@@ -0,0 +1,37 @@
+{ lib }:
+let
+  inherit (import ./semver.nix { inherit lib; }) satisfiesSemver;
+
+  # Returns true if pythonVersion matches with the expression in pythonVersions
+  isCompatible = pythonVersion: pythonVersions: let
```
zimbatm: maybe as part of the refactoring, moving the `let` keyword on its own line would be a good thing. It would put the `let` and `in` keywords on the same indent.

default.nix
andir: I guess the same reasons you might want to pass in `poetry`.

for `lib`: Being able to pass in a new version explicitly. I've had more then one case where the current project was on some kind of older nixpkgs revision but I had access to a newer `lib`. Many `x2nix` tools are only developed with unstable or the latest release in mind. Being able to get your stuff working anyway is a great value.

Same for the `poetryLib`. I can imagine being in a situation where I have to customize some kind of behavior of the nix code and not just the derivation. The more control we can expose to users without hurting yourself the better IMO.

lib.nix
```diff
@@ -0,0 +1,67 @@
...
+  readTOML = path: builtins.fromTOML (builtins.readFile path);
+
+  #
+  # Returns the appropriate manylinux dependencies and string representation for the file specified
+  #
+  getManyLinuxDeps = f:
+    let
+      ml = pkgs.pythonManylinuxPackages;
+    in
+      if lib.strings.hasInfix "manylinux1" f then { pkg = [ ml.manylinux1 ]; str = "1"; }
```
FRidh: note manylinux{1,2010,2014} and such are lists of packages. Therefore, it should not be inserted in a list.

FRidh: if we on the other hand convert them to derivations that *propagate* the libs and autoPatchElfHook it would be ok again (do name it `pkgs` then).

FRidh:
haven't checked in detail but on a glance this looks good to me.

lib.nix
gilligan: That's alright - if you look at the way the result of `getManyLinuxDeps` is used it's actually sane..

PR #29, opened 2020-01-09
Add some more overrides for setuptools-scm

robertodr:


adisbladis:
I have decided to always add `setuptools_scm` as a build input. The cost of doing so is very low and makes poetry2nix "just work" in a lot more cases.

https://github.com/nix-community/poetry2nix/commit/31a2bf6216a2b25e86fa84c2e0cdd42e668ac243

robertodr:
Makes a lot more sense. Thanks for this really awesome project!

PR #49, opened 2020-02-18
Handle two-layer inclusion of "path" dependencies

purcell:
This setup seems to work with poetry but not poetry2nix. We'll have a look at solving it, because we need to set up our build deps (in a monorepo) this way, but any comments/pointers are very welcome.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -194,13 +194,7 @@ let
             python = py;
           };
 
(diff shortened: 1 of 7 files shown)


purcell:
Have pushed a tentative fix now... happy to have feedback if it's problematic.

Issue #50, opened 2020-02-23
Python version bounds for `pyproject.toml` are ignored -> Building poetry with Python 3.8 fails


Infinisil:
Trying to build poetry with Python 3.8 fails:
```
$ nix-build -E 'with import <nixpkgs> {}; poetry.override { python = python38; }' -I nixpkgs=channel:nixpkgs-unstable
error: value is null while a set was expected, at /nix/store/24kjp6k6piq42jxvzx80imdx0ysz6grz-nixexprs.tar.xz
  /pkgs/development/tools/poetry2nix/poetry2nix/overrides.nix:569:27 (use '--show-trace' to show detailed location information)
```

This fails at https://github.com/nix-community/poetry2nix/blob/cdb1bd83a31d7ae12dd5b5b1ea460c11c942be16/overrides.nix#L569

because `super.zipp` is `null`. This is because `zipp` has a marker only indicating support for `< 3.8`: https://github.com/nix-community/poetry2nix/blob/cdb1bd83a31d7ae12dd5b5b1ea460c11c942be16/pkgs/poetry/poetry.lock#L1345-L1346

But the real problem is that `zipp` is evaluated at all, because `importlib-metadata`, the only package that depends on `zipp`, should only be a dependency for python < 3.8: https://github.com/nix-community/poetry2nix/blob/cdb1bd83a31d7ae12dd5b5b1ea460c11c942be16/pkgs/poetry/pyproject.toml#L58

And that condition is ignored because poetry2nix only parses package names, not their values: https://github.com/nix-community/poetry2nix/blob/cdb1bd83a31d7ae12dd5b5b1ea460c11c942be16/default.nix#L166-L170

I believe that's the correct diagnosis for this failure, and I think the solution is to parse these version constraints to conditionally include dependencies based on python version.

Digging through poetry's source code I found these methods for handling these version constraints: [`convert_dependencies`](https://github.com/python-poetry/poetry/blob/96191ca4f7551243366bda030426995fce5c4b57/poetry/masonry/builders/sdist.py#L272-L327) and [`to_pep_508`](https://github.com/python-poetry/poetry/blob/96191ca4f7551243366bda030426995fce5c4b57/poetry/packages/dependency.py#L201-L240). I think these might need to be implemented in Nix, which might be pretty easy with PEP 508 already implemented.

adisbladis:
Fixed in #52

PR #52, opened 2020-02-23
Add support for python version constraints from pyproject.toml

adisbladis:
Pep508 markers alone are not always enough, see #50

This sadly introduces a minor breakage in `mkPoetryEnv` (you now need to pass `pyproject` to `mkPoetryEnv` as well).

cc @Infinisil @andir @gilligan

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -29,7 +29,8 @@ let
   # Returns an attrset { python, poetryPackages } for the given lockfile
   #
   mkPoetryPython =
-    { poetrylock
+    { pyProject
+    , poetrylock
     , poetryPkg
     , overrides ? [ defaultPoetryOverrides ]
     , meta ? {}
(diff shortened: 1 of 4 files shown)


default.nix
```diff
@@ -50,7 +51,34 @@ let
 
       # Filter packages by their PEP508 markers
       partitions = let
-        supportsPythonVersion = pkgMeta: if pkgMeta ? marker then (evalPep508 pkgMeta.marker) else true;
+
+        supportsPythonVersion = pkgMeta: let
```
gilligan: I'm fine with the change per se but I'd very much like for the code below to be split up and moved to `lib.nix`. I find that this stack of let bindings kind of hides the semantics.

```nix
# Returns true if ...
supportsPythonVersion = pkgMeta: pyProject:
```

Looking at `lib.nix` I also see that the identical `combine` binding is over there as well so this could also be extracted. So :+1: on all of this but would be great if it could be refactored a bit such that machinery like `combine` and the logic of `the whole check yielding true weren't all happening on the same level which makes this harder to parse and follow - for my silly brain anyhow :sweat_smile:

adisbladis: Yeah they were pretty much identical, I've unified the logic.

PR #53, opened 2020-02-23
Expose mkPoetryPackages

utdemir:
This function is a useful building block when the Python packages
generated are used for another purpose.


adisbladis:
I've just refactored a bit and renamed `mkPoetryPython` -> `mkPoetryPackages`.

Updated your PR accordingly.

PR #57, opened 2020-02-25
Drop `meta` attribute for mkPoetryPackages/mkPoetryEnv

gilligan:
The `meta` attribute is only relevant for `mkPoetryApplication`
and can safely be dropped elsewhere.


gilligan:
Looking at the CI output I can see the removal didn't change anything at all and cached results were used so i'm gonna go ahead and merge this one.

PR #61, opened 2020-02-26
Set `src` to `cleanPythonSources projectDir` by default

adisbladis:
Adds a default source filter when `projectDir` is used
    
Also use the same source filter for path deps.
    
The default source filter currently filters:
- Filters gitignore (if it exists, non-recursive)
- Filters pycache/pyc files
- Uses cleanSourceFilter to filter out .git/.hg, .o/.so, editor backup files & nix result symlinks

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -96,6 +96,29 @@ let
         [ pythonPackages.${drvAttr} or (throw "unsupported build system ${buildSystem}") ]
       );
 
+  /*
+  Provides a source filtering mechanism that:
+
+  - Filters gitignore (if it exists, non-recursive)
+  - Filters pycache/pyc files
+  - Uses cleanSourceFilter to filter out .git/.hg, .o/.so, editor backup files & nix result symlinks
+  */
+  cleanPythonSources = { src }: let
+    rootGitignore = src + "/.gitignore";
+    hasGitignore = builtins.pathExists rootGitignore;
+    pycacheFilter = name: type:
+      (type == "directory" && ! lib.strings.hasInfix "__pycache__" name)
+      || (type == "regular" && ! lib.strings.hasSuffix ".pyc" name)
+    ;
+  in
+    lib.cleanSourceWith {
+      filter = lib.cleanSourceFilter;
+      src = lib.cleanSourceWith {
+        filter = pkgs.nix-gitignore.gitignoreFilterPure pycacheFilter (lib.optional hasGitignore rootGitignore) src;
+        inherit src;
+      };
+    };
+
 in
 {
   inherit
@@ -105,5 +128,6 @@ in
     readTOML
     getBuildSystemPkgs
     satisfiesSemver
+    cleanPythonSources
     ;
 }
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ Creates a Python application using the Python interpreter specified based on the
 - **meta**: application [meta](https://nixos.org/nixpkgs/manual/#chap-meta) data (_default:_ `{}`).
 - **python**: The Python interpreter to use (_default:_ `pkgs.python3`).
 
+Other attributes are passed through to `buildPythonApplication`.
+
 #### Example
 ```nix
 poetry2nix.mkPoetryApplication {
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ The _poetry2nix_ public API consists of the following attributes:
 
 Creates a Python application using the Python interpreter specified based on the designated poetry project and lock files. `mkPoetryApplication` takes an attribute set with the following attributes (attributes without default are mandatory):
 
-- **src**: project source.
 - **projectDir**: path to the root of the project.
+- **src**: project source (_default_: `lib.cleanSource projectDir`).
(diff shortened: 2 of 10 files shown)


lib.nix
```diff
@@ -96,6 +96,26 @@ let
         [ pythonPackages.${drvAttr} or (throw "unsupported build system ${buildSystem}") ]
       );
 
+  # Provides a source filtering mechanism that tries to do the right thing as much as possible:
+  # - Filters gitignore (if it exists, non-recursive)
+  # - Filters pycache/pyc files
+  # - Uses cleanSourceFilter to filter out .git/.hg, .o/.so, editor backup files & nix result symlinks
+  cleanPythonSources = src: let
```
gilligan: In general this all looks good to me, I just first want to try this on some project(s) to see if everything works out as expected. Beyond that:

Filtering on `.gitignore` makes perfect sense and I can't think of any reason not to do that. As for `__pycache__` and `.pyc` files I wonder if there is any chance that a package might actually need those? What if such files are included for testing? (Just like this PR adds a `.pyc` file).

One answer to that is _"Well then you can just specify `src` if you don't want filtering"_. This is actually a bit odd. Filtering happens as an implicit consequence of specifying `src`:

```nix
# this will be filtered
mkPoetryApplication {
 projectDir = ./.;
}
```
```nix
# this will not be filtered
mkPoetryApplication {
 projectDir = ./.;
 src = ./.;
}
```

I think this needs to be more explicit. Not sure about the best approach from the top of my head though. No filtering by default but there is a `poetry2nix.filterSrc` ? Add a `doFilter ? true` attribute? Not sure, what do you think?

andir: You should never have to distribute the `.pyc` files or `__pycache__` files. They are just an interpretation of the code that removes a parsing step from the interpreter.

I think when you supply `src` you should still have full control of what gets passed in without some magic filtering happening (by default).

adisbladis: While I agree the current usage is a bit odd I'm not sure what the API would look like otherwise.
I agree with @andir that I expect `src` to be passed without magic filtering if passed explicitly.

adisbladis: I've updated the docs. Maybe just documenting the behaviour is sufficient?

gilligan:
I still think this is a wrong decision.

I think whatever users pass as src should always be handled as is and not fiddled with in intransparent ways.

Removing files that aren’t essential is useful but i don’t think it justifies breaking this separation of concerns.

Documenting everything is good but it doesn’t change this.

adisbladis:
Whatever the user passes as `src` is not touched, the filter is a new default when `src` is _not_ passed.

adisbladis:
I'd like to clarify which part is the wrong decision as I'd like to tag a release asap.

andir:
On 05:13 29.02.20, Tobias Pflug wrote:
> I think whatever users pass as src should always be handled as is and not fiddled with in intransparent ways.

As far as I can tell `src` is never touched when provided. The default
for src is based on the project configuration.

Filtering gitignored and pycache files in the *default case* (e.g. when
*no* `src` attribute was given) sounds okay to me. Do you disagree?

> Removing files that aren’t essential is useful but i don’t think it
> justifies breaking this separation of concerns.

This is mostly about the out of the box experience with poetry2nix. If
you use potry2nix would you rather just tell it where your project files
are or also pass explicit cleaning of the directories? Removing
gitignored files and pycache files sounds *very* common to me.

Would you be happier if the filtering would be behind a specific toggle?

I actually believe users will benefit from having this as a default as
there will be less they have to work with.

If you disagree I am happy to talk about that (on monday :)).

gilligan:
@adisbladis @andir I can only repeat what i have said before:

```
# this will be filtered
foo = mkPoetryApplication {
 projectDir = ./.;
};

# this will not be filtered
bar = mkPoetryApplication {
 projectDir = ./.;
 src = ./.;
};
```
The fact that filtering occurs on whether or not `src` is provided seems weird to me. 

I think in general filtering is something that is not the responsibility of the library/builder which should just rely on whatever is being specified - and I don't think it matters if this happens through an indirection (`projectDir`) or not.

Whenever local sources are being build additional filtering will probably have to be applied, unless you want to extend the filtering to `*.o *.swp *.bak *.git ...` which, theoretically, could be beneficial to the user as well.

So anyway, i would prefer for filtering to be applied but considering it has already been merged and both of you are in favor i don't  quite dislike it enough to argue any further and will just let it be.

adisbladis:
> Whenever local sources are being build additional filtering will probably have to be applied, unless you want to extend the filtering to *.o *.swp *.bak *.git ... which, theoretically, could be beneficial to the user as well.

This can be achieved since we're using `lib.cleanSourceWith` that composes, so you could do (this should go in the future development docs):
``` nix
lib.cleanSourceWith {
  filter = name: type: ! lib.hasSuffix ".bak" name;
  src = poetry2nix.cleanPythonSources {
    src = ./.;
  }
}
```

Also: I'm not opposed to extending the default source filter further.

> So anyway, i would prefer for filtering to be applied but considering it has already been merged and both of you are in favor i don't quite dislike it enough to argue any further and will just let it be.

Thanks. I wanted to make sure there was no misunderstanding but instead a difference of opinion.

PR #62, opened 2020-02-26
Add overrides for fancycompleter and pylint

Infinisil:
Needed these to work. The `versioning` override for fancycompleter took a while to figure out.

I'm not entirely sure how to test these, but I can say that these overrides work in the project I'm working on.

This PR is sponsored by [Niteo](https://niteo.co/) :sparkles:

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -96,6 +96,16 @@ self: super:
     }
   );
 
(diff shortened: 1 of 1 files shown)


adisbladis:
Thanks!

PR #63, opened 2020-02-27
Add development tool to generate flame graphs for nix-instantiate

adisbladis:


shell.nix
```diff
@@ -1,11 +1,17 @@
 let
-  pkgs = import (fetchTarball { url = "channel:nixpkgs-unstable"; }) {};
-  poetry2nix = import ./. { inherit pkgs; inherit poetry; };
-  poetry = pkgs.callPackage ./pkgs/poetry { python = pkgs.python3; inherit poetry2nix; };
+  pkgs = import <nixpkgs> {
```
gilligan: Is switching from `channel:nixpkgs-unstable` to `<nixpkgs>` intentional here? I think using `nixpkgs-unstable` has a better chance of working as poetry2nix pretty regularly depends on recent nixpkgs revisions.

gilligan: actually, how about
```nix
{ pkgs ? import (fetchTarball { url = "channel:nixpkgs-unstable"; }) {} }:
```
i like being able to specify nixpkgs when invoking the shell.

adisbladis: It was intentional but you're right.
It's especially true for `nixpkgs-fmt` which have broken it's formatting once before and we'd get different results on different channels.

adisbladis: The problem with `{ pkgs ? ... }: ..` is that there isn't a nice way to allow passing `pkgs` that would always use the overlay.

Maybe it's fine anyway though? A user wanting to pass `pkgs` would have to take care to use the overlay themselves.

adisbladis: Implemented, it's up to a user wanting to pass a custom `pkgs` to make sure what they're passing makes sense.

tools/default.nix
```diff
@@ -0,0 +1,46 @@
...
+    srcPath = builtins.toString ../.;
+
+  in
+    pkgs.writeScriptBin "poetry2nix-flamegraph" ''
+      #!${pkgs.runtimeShell}
+      export PATH=${runtimeDeps}:$PATH
+
+      workdir=$(mktemp -d)
+      function cleanup {
+        rm -rf "$workdir"
```
gilligan: I tend to get nervous and `shellcheck` yells at me when I try to write `rm -rf $SOMETHING`. I know the scope is small here but it would be nice to rewrite this to something with a `/tmp` prefix: `rm -rf /tmp/$whatever`

gilligan:
This is probably beyond the scope of this PR but I wonder: is there any meaningful way to incorporate this into the CI ? Does gitlab CI support some sort of artifact or so such that the flamegraph could be visible?

adisbladis:
Added a flamegraph artifact: https://gitlab.com/nix-community/poetry2nix/-/jobs/451932135/artifacts/browse/artifacts/

tools/default.nix
adisbladis: I don't know how else this would be implemented in a shell-script so I'll leave it for now. I'm also not a fan of the pattern and would much rather have something like python's `tempfile` which can clean things up when exiting the context manager.

PR #65, opened 2020-02-27
Fix pylint override for more recent versions

gilligan:
The `setup.py` in a more recent version of `pylint` changed. This adds another substitution so that both cases are covered.

Note that `substituteInPlace` doesn't fail if no match is found, it just prints a warning.


adisbladis:
Fixed by adding as a build input in another PR

PR #72, opened 2020-03-12
Do not leak projectDir into expressions

stolyaroleh:
Fixes caching issues.


adisbladis:
Doh! No-brainer. Thanks!

PR #78, opened 2020-03-30
overrides: add python-jose

rissson:
`python-jose` lists `pytest-runner` in its `setup_requires` even though it is not required for setup. The package fthus ailed to build.


adisbladis:
Thanks!

PR #79, opened 2020-03-31
Automatically build most packages without passing format param

adisbladis:
We should never have to use `format = "setuptools"` since pip can
_also_ handle setuptools packages.

This uses a custom pipBuildHook that removes pyproject.toml in case
there is also a setup.py in the same sources.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -102,7 +102,7 @@ let
                     # The canonical name is setuptools-scm
                     setuptools-scm = super.setuptools_scm;
 
-                    inherit (hooks) removePathDependenciesHook poetry2nixFixupHook;
+                    inherit (hooks) pipBuildHook removePathDependenciesHook poetry2nixFixupHook;
                   }
             )
             # Null out any filtered packages, we don't want python.pkgs from nixpkgs
--- a/hooks/default.nix
+++ b/hooks/default.nix
@@ -2,9 +2,12 @@
 , callPackage
 , makeSetupHook
 , yj
+, wheel
+, pip
 }:
 let
   pythonInterpreter = python.pythonForBuild.interpreter;
+  pythonSitePackages = python.sitePackages;
 in
 {
 
@@ -19,6 +22,17 @@ in
           pyprojectPatchScript = "${./pyproject-without-path.py}";
         };
       } ./remove-path-dependencies.sh
+  ) {};
+
+  pipBuildHook = callPackage (
+    { pip, wheel }:
+      makeSetupHook {
+        name = "pip-build-hook.sh";
+        deps = [ pip wheel ];
+        substitutions = {
+          inherit pythonInterpreter pythonSitePackages;
+        };
+      } ./pip-build-hook.sh
   ) {};
 
   poetry2nixFixupHook = callPackage (
--- a/hooks/pip-build-hook.sh
+++ b/hooks/pip-build-hook.sh
@@ -1,0 +1,50 @@
+# Setup hook to use for pip projects
+echo "Sourcing pip-build-hook"
+
+pipBuildPhase() {
+    echo "Executing pipBuildPhase"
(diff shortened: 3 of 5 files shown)


adisbladis:
cc @FRidh

FRidh:
> This uses a custom pipBuildHook that removes pyproject.toml in case
there is also a setup.py in the same sources.

It's perfectly fine for `pyproject.toml` to specify `setuptools` needs to be used.

Pip is just the front-end, a back-end (setuptools, flit, poetry, meson, ...) is always going to be needed.

As I wrote on IRC, support for setuptools arguments needs to be added (e.g. arguments to `build_ext`).

FRidh:
In the long-run (or maybe already soon) we can get rid of `format` and make `pyproject` the default. If people have packages that do not support PEP517/518/`pyproject.toml`, then they can still add one of these hooks.

FRidh:
Also, this works because you always propagate `setuptools`. It's pragmatic but incorrect.

adisbladis:
> In the long-run (or maybe already soon) we can get rid of format and make pyproject the default

`format = "pyproject"` is, I think, wrongly named. A better name would have been `format = "pip"` as that's what the option actually implies. 
`pipBuildHook` works just as well on setuptools packages but will default to using `pyproject.toml` if it's available, potentially adding more required build inputs than using plain `setup.py` because of `build-system`.

I think we can _already_ change the default (and maybe add a new `format = "pip"` that does the same as the current `pyproject` param).
I don't expect to see any breakage from this (provided `pipBuildHook` would _also_ propagate setuptools).

> It's perfectly fine for pyproject.toml to specify setuptools needs to be used as back-end.
> Pip is just the front-end, a back-end (setuptools, flit, poetry, meson, ...) is always going to be needed.

I want to remove https://github.com/nix-community/poetry2nix/pull/79/files#diff-28614fffa371111fed2fbc35ca55fb67R9 and use the nixpkgs `pipBuildHook` if we could get Poetry to add build time dependencies to the lock file.

> Also, this works because you always propagate setuptools. It's pragmatic but incorrect.

It may be incorrect, but I want `poetry2nix` to "just work" to the largest extent possible which makes this a requirement.
Hopefully this is something we can address through adding better meta data to `poetry.lock`.

FRidh:
> format = "pyproject" is, I think, wrongly named. A better name would have been format = "pip" as that's what the option actually implies.

That depends on your interpretation. To me `pip` is an implementation detail. We just say, use PEP 517/518/pyproject.toml. How that is done, is up to `buildPython*`. The whole `format` thing predates PEP 517/PEP518/pyproject.toml. In the long run we should get rid of it and just let users include the relevant setup hook . That typically means PEP 517/518/pyproject.toml or wheel or any of the legacy ones (such as setuptools).

> pipBuildHook works just as well on setuptools packages but will default to using pyproject.toml if it's available, potentially adding more required build inputs than using plain setup.py because of build-system.

In your implementation of `pipBuildHook` that is indeed how it works. While it may be *convenient* it conflates two different ways and also requires `setuptools` to *always* be present, even when `setuptools` is not used as back-end. Remember, there are also still *other* build methods that predate PEP 517/518 although granted they're few and not really utilized out there nor in Nixpkgs; the few flit packages we have I think use pyproject already.

> It may be incorrect, but I want poetry2nix to "just work" to the largest extent possible which makes this a requirement.

Right, I understand that. It will cover 99.99% of the cases for you. And you can indeed take that approach with `poetry2nix`, but I am against doing this in the main builder.

But still, 

> This uses a custom pipBuildHook that removes pyproject.toml in case
there is also a setup.py in the same sources.

is incorrect. Setuptools projects that start supporting PEP 517/518 will have both a `pyproject.toml` *and* a `setup.py` (and possibly a `setup.cfg`).

PR #87, opened 2020-04-15
Add canonicalisation of module names

### LINKED ISSUE #84: Missing dependency?
`error: attribute 'pyramid_retry' missing, at /nix/store/bj56vllyvxfn7y2chm6l53rimyh9jmms-source/pkgs/development/tools/poetry2nix/poetry2nix/mk-poetry-dep.nix:132:28`

I'm not entirely sure what's going on, pyramid_retry only appears in `poetry.lock`. It seems like something depends on it before `pythonPackages.pyramid_retry` is created?

I can put together an example repo if that would help.

cc @adisbladis

adisbladis:
This shouldn't be required and I think that these are actually packaging bugs and not something that poetry2nix should really be concerned with.
With that said, if it works in Poetry the same lock file should also work in Poetry2nix.

Closes #84

cc @Infinisil @domenkozar

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,7 +4,7 @@
 , poetryLib ? import ./lib.nix { inherit lib pkgs; }
 }:
 let
-  inherit (poetryLib) isCompatible readTOML;
+  inherit (poetryLib) isCompatible readTOML moduleName;
 
   # Poetry2nix version
   version = "1.7.1";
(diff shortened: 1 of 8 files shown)


domenkozar:
I would be also OK if poetry2nix would fail in this case and advise how to proceed.

These are upstream bugs that are harder to solve and get across, but at least it's clear what the problem is.

adisbladis:
> I would be also OK if poetry2nix would fail in this case and advise how to proceed.

I don't think we can reasonably ask end users to proceed as this is an upstream packaging bug. I guess we could ask the user of poetry2nix to add an "override" that aliases one name to the other but it feels more hacky than a simple string transform.

purcell:
> if it works in Poetry the same lock file should also work in Poetry2nix

Thanks for this change: even if it shouldn't really be necessary, I agree with this ☝️

PR #89, opened 2020-04-16
Fix wandb build by setting configparser version

utdemir:
This PR adds a failing test case for the `wandb` package:                                                                                                                                                            
                                                                                                                                                                                                                     
```                                                                                                                                                                                                                  
ERROR: Could not find a version that satisfies the requirement configparser>=3.8.1 (from wandb==0.8.28) (from versions: none)                                                                                        
ERROR: No matching distribution found for configparser>=3.8.1 (from wandb==0.8.28)                                                                                                                                   
```                                                                                                                                                                                                                  
                                                                                                                                                                                                                     
This happens despite the fact that `wandb` declares `configparser` dependency. We also tried override it adding `configparser` to the build inputs but it didn't help.                                               
                                                                                                                                                                                                                     
From bisecting the poetry2nix repository, this started to happen after this change: https://github.com/nix-community/poetry2nix/commit/b9221f61b581f8906bf99d8657945f27e9cca1e8.                                     
                                                                                                                                                                                                                     
We'd be happy to fix it if you have any pointers.                                                                                                                                                                    
                                                                                                                                                                                                                     
/cc @gilligan

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -51,6 +51,10 @@ self: super:
       buildInputs = old.buildInputs ++ [
         self.toml
       ];
+
+      postPatch = ''
+        substituteInPlace setup.py --replace 'setuptools.setup()' 'setuptools.setup(version="${old.version}")'
+      '';
     }
   );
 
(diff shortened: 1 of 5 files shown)


adisbladis:
Thanks for pointing this out!

purcell:
Awesome, thanks @adisbladis!

PR #95, opened 2020-04-29
Introduce `dependencyEnv` passthru to `mkPoetryApplication`

Infinisil:
A combination of `mkPoetryApplication` and `mkPoetryEnv`: An environment with a python interpreter, all dependencies and the application built from pyproject.toml as well. In addition, all binaries know about the python modules of the application. This is useful if your application doesn't have a binary on its own, but relies on dependencies binaries to call its modules.

A test case might be a good idea, but I can say that this works very well with the project I've been testing this with locally.

Ping @adisbladis @energizah

Edit: This PR is newly sponsored by [Niteo](https://niteo.co/)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -229,9 +229,15 @@ let
 
             passthru = {
               python = py;
-              dependencyEnv = py.buildEnv.override {
-                extraLibs = getDeps "dependencies" ++ [ app ];
-              };
+              dependencyEnv =
+                (lib.makeOverridable
+                  ({ app, ... }@attrs:
+                    let
+                      args = builtins.removeAttrs attrs [ "app" ] // {
+                        extraLibs = [ app ];
+                      };
+                    in
+                    py.buildEnv.override args)) { inherit app; };
             };
 
             meta = meta // {
--- a/default.nix
+++ b/default.nix
@@ -150,7 +150,13 @@ let
     in
     py.python.withPackages (_: py.poetryPackages);
 
-  /* Creates a Python application from pyproject.toml and poetry.lock */
+  /* Creates a Python application from pyproject.toml and poetry.lock
+
+     The result also contains a .dependencyEnv attribute which is a python
+     environment of all dependencies and this apps modules. This is useful if
+     you rely on dependencies to invoke your modules for deployment: e.g. this
+     allows `gunicorn my-module:app`.
+  */
   mkPoetryApplication =
     { projectDir ? null
     , src ? poetryLib.cleanPythonSources { src = projectDir; }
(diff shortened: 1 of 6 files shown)


Infinisil:
Thinking about this again, I think this could be implemented with just an `includeDependencies` parameter to `mkPoetryApplication`

Infinisil:
Now implemented this as part of `mkPoetryApplication` with a `dependencyEnvironment` parameter. Also added a test case which illustrates the use-case.

default.nix
```diff
@@ -202,34 +206,44 @@ let
...
+              inherit (py.meta) platforms;
+              license = getLicenseBySpdxId (pyProject.tool.poetry.license or "unknown");
+            };
 
-        }
-      );
+          }
+        );
+      appEnv = py.buildEnv.override {
+        extraLibs = poetryPython.poetryPackages ++ [ app ];
```
adisbladis, line 241: This is a bit broad. In the case of `mkPoetryApplication` we don't want to include `dev-dependencies` in this call to `buildEnv`.

Infinisil: Agreed, now fixed by having `mkPoetryPackages` expose a separate attribute for runtime-specific dependencies

Infinisil: Oh duh, there's `getDeps "dependencies"` I can use instead. Doing that now

default.nix
```diff
@@ -202,34 +206,44 @@ let
...
+            };
 
-        }
-      );
+          }
+        );
+      appEnv = py.buildEnv.override {
+        extraLibs = getDeps "dependencies" ++ [ app ];
+      };
+    in
```
adisbladis, line 243: I wonder if it isn't better to stick the env in a passthru attribute?

That way you can package an application using `mkPoetryApplication` as usual and get the desired environment in for example a NixOS module.

Infinisil: As we discussed on IRC, a passthru seems better overall, changed it to that

tests/dependency-environment/default.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  # Test support for overriding the app passed to the environment
+  overriden = (
+    app.overrideAttrs
+      (old: {
+        name = "${old.pname}-overriden-${old.version}";
+      })
+  );
+  depEnv = app.dependencyEnv.override {
+    app = overriden;
```
cole-h, line 18:
```suggestion
  overridden = (
    app.overrideAttrs
      (old: {
        name = "${old.pname}-overridden-${old.version}";
      })
  );
  depEnv = app.dependencyEnv.override {
    app = overridden;
```

:D

gilligan:
damn, i'm too late. I was too busy. There is no documentation for this at all. Can you please, please add documentation for this in the README @Infinisil ?

Issue #59, opened 2020-02-26
Write Tests for all overrides in overrides.nix


gilligan:
Ideally all fixes in [overrides.nix](./overrides.nix) should be covered by a test for the respective package.This might be true for some of the packages but certainly not all of them.

adisbladis:
While ideal this is unrealistic.

PR #97, opened 2020-05-04
Fix isort override

purcell:
Sorry, I'm an idiot. Fixes mistake in #96.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -174,7 +174,7 @@ self: super:
   isort = super.isort.overridePythonAttrs
     (
       old: {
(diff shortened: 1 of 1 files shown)


purcell:
🙏  Thanks!

adisbladis:
We are all idiots sometimes. Me too for merging ^_^

adisbladis:
Note to self: We need tests for overrides.

purcell:
> We need tests for overrides.

Yep, there's #59. I guess one approach might be to programmatically build a `pyproject.toml` that depends on all the overrides, then lock it and build it. But it might get pretty big and slow.

PR #100, opened 2020-05-07
add override for panel, imagecodecs

tbenst:
Note that imagecodecs override depends on https://github.com/NixOS/nixpkgs/pull/87205.

Also, should such large overrides go into this single file? I could see this getting out of hand quickly as it somewhat duplicates nixpkgs for complicated packages...

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -163,6 +163,64 @@ self: super:
       }
     );
 
+  imagecodecs = super.imagecodecs.overridePythonAttrs
+    (
+      old: {
+        patchPhase = ''
+          substituteInPlace setup.py \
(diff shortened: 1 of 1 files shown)


adisbladis:
> Also, should such large overrides go into this single file?

I think we're about reaching the breaking point where the single file approach is no longer viable and this should be split up.
For this PR let's stick with the single file and let me clean up afterwards.

adisbladis:
I made two fixups:

- You were mixing `nativeBuildInputs` and `buildInputs` when creating the overriden `buildInputs`.
- The use of `with` is not allowed in poetry2nix, so I've explicitly added `pkgs.` before each input.

PR #101, opened 2020-05-08
Introduce `editablePackageSources` to `mkPoetryEnv`

Infinisil:
This introduces a new parameter to `mkPoetryEnv` which can be used to specify sources for additional python packages that should be available in the resulting environment, loading directly from the mutable source directory. This allows you to call binaries that need to load python modules (like `python -m` or `gunicorn`) for development.

This `editablePackageSources` parameter *could* be read from the [packages](https://python-poetry.org/docs/pyproject/#packages) entry in `pyproject.toml` (or inferred if it doesn't exist), and that's what `poetry install` does, but this would be rather complex to implement correctly, so passing it manually seems to make more sense to me.

The implementation of this works the same way as `poetry install` works: `.pth` files (somewhat documented [here](https://docs.python.org/3.8/library/site.html)) are created in the environments site-packages directory, which reference the mutable directory and lets python know to find the modules there.

This PR is sponsored by [Niteo](https://niteo.co/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,13 +62,14 @@ $ result/bin/gunicorn web:app
 Note: If you need to perform overrides on the application, use `app.dependencyEnv.override { app = app.override { ... }; }`. See [./tests/dependency-environment/default.nix](./tests/dependency-environment/default.nix) for a full example.
 
 ### mkPoetryEnv
-Creates an environment that provides a Python interpreter along with all dependencies declared by the designated poetry project and lock files. `mkPoetryEnv` takes an attribute set with the following attributes (attributes without default are mandatory):
+Creates an environment that provides a Python interpreter along with all dependencies declared by the designated poetry project and lock files. Also allows package sources of an application to be installed in editable mode for fast development. `mkPoetryEnv` takes an attribute set with the following attributes (attributes without default are mandatory):
 
 - **projectDir**: path to the root of the project.
 - **pyproject**: path to `pyproject.toml` (_default_: `projectDir + "/pyproject.toml"`).
 - **poetrylock**: `poetry.lock` file path (_default_: `projectDir + "/poetry.lock"`).
 - **overrides**: Python overrides to apply (_default_: `[defaultPoetryOverrides]`).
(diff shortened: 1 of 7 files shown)


Infinisil:
Hm turns out this doesn't quite work in all cases.
- Using a `pth` file makes `import`'s work by having additional paths in `sys.path`
- But [`pkg_resources.require()`](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-workingset-methods) which is apparently often used to support plugins, doesn't work with just that. To support this an `.egg-info` directory needs to be generated, and an `.egg-link` file pointing to the `.egg-info` needs to be installed in the site packages directory. See [here](https://discuss.python.org/t/specification-of-editable-installation/1564) for a bit more info.

Now poetry does this by [generating a `setup.py`](https://github.com/python-poetry/poetry/blob/621f698ab7fa46b4109a6812a11c325dd2222746/poetry/masonry/builders/sdist.py#L104-L196) (or using the one from the project if it exists), and then [calling `pip install -e /path/to/setup.py`](https://github.com/python-poetry/poetry/blob/621f698ab7fa46b4109a6812a11c325dd2222746/poetry/masonry/builders/editable.py#L15-L49) (which is a bit nasty). I believe `python /path/to/setup.py develop` is used underneath.

But after some digging on the `.egg-info` format (e.g. the docs [here](https://setuptools.readthedocs.io/en/latest/formats.html)) it looks like we can easily generate that directory statically. The only thing that's required is a `PKG-INFO` file within it, containing basic metadata as described in [PEP 314](https://www.python.org/dev/peps/pep-0314/).

In addition, an `entry_points.txt` file can be present, which is just a direct conversion from the [`tool.poetry.plugins` key](https://python-poetry.org/docs/pyproject/#plugins) to INI.

Both of these files can therefore relatively easily be generated from the `pyproject.toml`. I'll try to implement this soon.

Infinisil:
I implemented this now in https://github.com/nix-community/poetry2nix/pull/110

PR #103, opened 2020-05-10
Split wheel overrides from sdist overrides

adisbladis:
This is to improve the user experience when prefering wheels globally.


adisbladis:
This approach turned out too messy and unmaintainable I think... 
I need to go back to the drawing board and reconsider approaches.

adisbladis:
This PR is not viable, even if I still haven't found a better approach. Closing for now.

PR #110, opened 2020-05-23
Editable eggs

Infinisil:
In https://github.com/nix-community/poetry2nix/pull/101, `editablePackageSources` was introduced, but later [I discovered](https://github.com/nix-community/poetry2nix/pull/101#issuecomment-629870767) that this doesn't work for [`pkg_resources.require()`](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-workingset-methods), which needs an installed egg to find the package.

With this change, this is fixed by making `mkPoetryEnv` install a very basic static egg.

In the first commit, a failing test is added, showcasing what didn't work previously, whereas the second commit makes it work.

This PR is sponsored by [Niteo](https://niteo.co/)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -150,20 +150,14 @@ let
           inherit pyproject poetrylock overrides python pwd preferWheels;
         }
       );
-      editablePackages = lib.mapAttrsToList
-        (name: src:
-          # Creates a "fake" python module that just points to the editable source directory
-          # See https://docs.python.org/3.8/library/site.html for info on such .pth files
-          py.python.pkgs.toPythonModule (pkgs.runCommandNoCC "${name}-editable"
-            { } ''
-            mkdir -p "$out/${py.python.sitePackages}"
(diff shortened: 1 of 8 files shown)


default.nix
```diff
@@ -150,20 +150,52 @@ let
...
-          # See https://docs.python.org/3.8/library/site.html for info on such .pth files
-          py.python.pkgs.toPythonModule (pkgs.runCommandNoCC "${name}-editable"
-            { } ''
-            mkdir -p "$out/${py.python.sitePackages}"
-            echo "${toString src}" > "$out/${py.python.sitePackages}/${name}.pth"
-          ''
-          )
-        )
-        editablePackageSources;
+
```
adisbladis, line 153: Could you split this to a separate file?

Infinisil: Like this?

adisbladis: Nice, thanks!

PR #119, opened 2020-06-10
Make pyarrow override compatible with newer nixpkgs

utdemir:
Since the nixpkgs change https://github.com/NixOS/nixpkgs/commit/f149c7030a7#diff-b7d8c4fa8c1a2ad2832a889247b62e3bL3, `arrow-cpp` package expects an argument named `python3` instead of just `python`. This commit adds a compatibility fix so that the package can also be used with a newer nixpkgs.


adisbladis:
Thanks!

PR #135, opened 2020-07-14
Use poetry.masonry.api build-backend by default

Infinisil:
Because `poetry build` works too without having any `build-backend` specified.

Not doing this can cause problems for the final application, a symptom of which seems to be that the package metadata isn't detected, leading to UNKNOWN eggs in the python environment.

An alternative would be to throw an error or a warning, pushing people to adding a `build-backend` section, but since peotry doesn't do that, poetry2nix shouldn't do it too imo.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -154,12 +154,10 @@ let
     , pyProject
     }:
     let
-      buildSystem = lib.attrByPath [ "build-system" "build-backend" ] "" pyProject;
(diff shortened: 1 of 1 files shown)


Infinisil:
Never mind, this doesn't fix the problem, you actually need to have such a section for the UNKNOWN problem to go away

PR #137, opened 2020-07-21
Wrap poetry2nix in a scope to make internals overridable

adisbladis:
Sometimes when it can be convenient to create a custom instance of `poetry2nix` with a different set of default overrides.

#### Example
```nix
let
  # self & super refers to poetry2nix
  p2nix = poetry2nix.overrideScope' (self: super: {

    # pyself & pysuper refers to python packages
    defaultPoetryOverrides = super.defaultPoetryOverrides.extend (pyself: pysuper: {

      my-custom-pkg = super.my-custom-pkg.overridePythonAttrs (oldAttrs: { });

    });

  });

in
p2nix.mkPoetryApplication {
  projectDir = ./.;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,35 @@ p2nix.mkPoetryApplication {
 }
 ```
 
+or as a [nixpkgs overlay](https://nixos.org/nixpkgs/manual/#chap-overlays):
+```nix
+let
+  pkgs = import <nixpkgs> {
+    overlays = [
+      # self & super refers to nixpkgs
+      (self: super: {
+
+        # p2self & p2super refers to poetry2nix
+        poetry2nix = super.poetry2nix.overrideScope' (p2nixself: p2nixsuper: {
+
+          # pyself & pysuper refers to python packages
+          defaultPoetryOverrides = p2nixsuper.defaultPoetryOverrides.extend (pyself: pysuper: {
+
+            my-custom-pkg = super.my-custom-pkg.overridePythonAttrs (oldAttrs: { });
+
+          });
+
+        });
+      })
+
+    ];
+  };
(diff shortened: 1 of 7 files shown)


grahamc:
I've heard confirmation that this makes it possible to by-default override packages which bundle certificate authorities, solving their problem well. Thank you a lot!

grahamc:
By the way, here is an example of that use case. Where `pkgs.cacerts` has been overridden:

```nix
let
  pkgs = import <nixpkgs> {};
in
self: super:
  {
    certifi = super.certifi.overridePythonAttrs (
      { postPatch ? "", ... }: {
        postPatch = postPatch + ''
          cp -f ${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt certifi/cacert.pem
        '';
      }
    );
    boto = super.boto.overridePythonAttrs (
      { postPatch ? "", ... }: {
        postPatch = postPatch + ''
          cp -f ${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt boto/cacerts/cacerts.txt
        '';
      }
    );
    botocore = super.botocore.overridePythonAttrs (
      { postPatch ? "", ... }: {
        postPatch = postPatch + ''
          cp -f ${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt botocore/cacert.pem
        '';
      }
    );
  }
```

though I wonder now if these overrides could actually go upstream...

adisbladis:
> though I wonder now if these overrides could actually go upstream...

I think that's preferable if possible.

asymmetric:
@grahamc IIUC the overrides above haven't been upstreamed, right?

Issue #139, opened 2020-07-22
Dependency issues with ffmpeg-python


mtn:
There's an error while building a project with `ffmpeg-python` as its only dependency. It looks like it tries to make a network request while build? Here's the error:

```
    ERROR: Command errored out with exit status 1:
     command: /nix/store/j3q3f6nmaif7lmqdsp8fg2hxxrd4w1gh-python3-3.7.6/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/build/pip-req-build-d2bwucs0/setup.py'"'"'; __file__='"'"'/build/pip-req-build-d2bwucs0/setup.
py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /build/pip-req-build-d2bwucs0/pip-egg-info
         cwd: /build/pip-req-build-d2bwucs0/
    Complete output (38 lines):
    WARNING: The directory '/homeless-shelter/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo,
 you may want sudo's -H flag.
    WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7ffff5ef58d0>: Failed to estab
lish a new connection: [Errno -2] Name or service not known')': /simple/pytest-runner/
    WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7ffff5f0e750>: Failed to estab
lish a new connection: [Errno -2] Name or service not known')': /simple/pytest-runner/
    WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7ffff5f0e950>: Failed to estab
lish a new connection: [Errno -2] Name or service not known')': /simple/pytest-runner/
[...]

PR #140, opened 2020-07-22
Add override for ffmpeg-python

mtn:
Adds `pytest-runner` as a build input, handling #139


adisbladis:
I removed the `rec` as the set was not accessing any attributes on itself.

PR #147, opened 2020-08-12
Added openssl dependency to psycopg2 on Darwin.

### LINKED ISSUE #146: Psycopg2 on mac needs openssl dependency
This issue only shows up when one wants to use psycopg2 on a mac. That dependency is already in nixpkgs, but it is missing in poetry2nix. I will submit a PR with the change.

takeda:
This should fix #146 

closes #146


adisbladis:
Thanks!

PR #149, opened 2020-08-13
overrides: Matplotlib 3.3.0 now downloads/builds it's own freetype by default

adisbladis:
Override to use system (nix) freetype.


adisbladis:
This is the PR that introduced the change that broke our override: https://github.com/matplotlib/matplotlib/pull/15476.

PR #158, opened 2020-08-25
Override for dictdiffer

rskew:
I've tested this out and it works. I'm happy to add a formal test to the repo too, but not sure exactly what the right way to test this is.


adisbladis:
Thanks!

Issue #160, opened 2020-08-26
does poetry2nix support {git = ... tag = ...} packages ?


teto:
with a shell.nix similar to 
```
myPythonEnv = poetry2nix.mkPoetryEnv {
    projectDir = ./.;
  };
```
I get upon running nix-shell
```
fatal: impossible de trouver la référence distante refs/heads/v6.3.0
error: --- Error ------------------------------------------------------------------------------------------------------------- nix-shell
program 'git' failed with exit code 128
-------------------------------------------------------------- show-trace --------------------------------------------------------------
trace: while evaluating the attribute 'src.name'
```
`fatal: *can't find remote reference* refs/heads/v6.3.0` (using nixos-unstable c59ea8b8a0e7f927e7291c14ea6cd1bd3a16ff38)
```
[[package]]
category = "main"
name = "custom-sdk"
optional = false
python-versions = "*"
version = "6.3.0"

[package.dependencies]
pandas = ">=0.25.0"
retrying = ">=1.3.3"
tqdm = ">=4.38.0"

[package.source]
reference = "660a747c9ef2a749c88fd87f4e009dd514352992"
type = "git"
url = "https://REDACTED/custom-sdk.git"

```
In a local checkout of the git repo, the tags are in .git/refs/tags and not in .git/refs/heads. I am not sure how git handles tags/refs but I am concerned the { git = , tag } pattern in pyproject.toml is not supported yet in poetry2nix ? or is it something else ? 

the pyproject.toml contains
`
custom-sdk = {git = 'https://redacted/custom-sdk.git', tag = 'v6.3.0'}
`

adisbladis:
This should be supported: https://github.com/nix-community/poetry2nix/blob/10e417df85ec26d9f808c9595557f627e3714336/mk-poetry-dep.nix#L157.

Are we somehow constructing the `builtins.fetchGit` call wrongly?
Could you try to manually call `builtins.fetchGit` and see what you can get working?

teto:
poetry2nix looks for the ref in  refs/heads/v6.3.0 when it really is in  refs/tags/v6.3.0:
```
$ git show-ref                                                                                                         
ebcafd9e6337ade03a88cdf36eb59d480c9d60d3 refs/heads/dev
ebcafd9e6337ade03a88cdf36eb59d480c9d60d3 refs/remotes/origin/HEAD
ebcafd9e6337ade03a88cdf36eb59d480c9d60d3 refs/remotes/origin/dev
[...]

PR #167, opened 2020-08-31
fix: poetry wouldn't work when using a tag

teto:
fixes https://github.com/nix-community/poetry2nix/issues/160

I copy/pasted the patch from my fork of nixpkgs here and could not properly test.

### DIFF
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -153,9 +153,13 @@ pythonPackages.callPackage
         if isGit then (
           builtins.fetchGit {
             inherit (source) url;
-            rev = source.reference;
-            ref = sourceSpec.branch or sourceSpec.rev or sourceSpec.tag or "HEAD";
-          }
+          } // (
+            if source.reference != null then {
+              rev = source.reference;
+            } else {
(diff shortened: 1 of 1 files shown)


teto:
@adisbladis I am trying to write tests for this to convince you (and also fixed an issue in my previous PR).
I've been running nix-build tests/default.nix but if fails with unknown package cppy ?
upon entering nix-shell, nixpkgs is overriden to one that doesn't seem valid  but if I restore nixpkgs to my local checkout with `export NIX_PATH=nixpkgs=/home/teto/nixpkgs` (a nixos-unstable fork) tests work again. Not sure what's wrong, I never used niv.

teto:
while writing the test https://github.com/nix-community/poetry2nix/pull/171 for this, it seemed like poetry works already fine so don't merge this yet. I wonder if the error I had was not a sideeffect of https://github.com/python-poetry/poetry/issues/2062 in fact.

teto:
doesn't seem needed after all, I wonder if I got mystified by a mismatched locked file and pyproject.toml. Closing.

PR #168, opened 2020-09-01
improving flake experience

teto:
I've been experimenting with flakes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,8 +1,6 @@
 {
   description = "Poetry2nix flake";
 
-  edition = 201909;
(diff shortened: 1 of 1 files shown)


adisbladis:
I don't think there is any point in poetry2nix adding a `flake.lock`.

teto:
what about removing 'edition' xD (I can drop the lock commit) ?

adisbladis:
Removing `edition` is a :+1:.

> I can drop the lock commit

Please do.

teto:
done

PR #173, opened 2020-09-14
Extra overrides

xwvvvvwx:
I needed these overrides to make my project build with `poetry2nix`.

I was a little confused why I needed to add the overrides for `setuptools-markdown` in e.g. `eth-hash`, because it's already present in the `nixpkgs` [package definition](https://github.com/NixOS/nixpkgs/blob/097cf628d9aa3c9a405f63f128ca6e222570e17b/pkgs/development/python-modules/eth-hash/default.nix), but maybe I'm just misunderstanding how `poetry2nix` does it's magic?

Awesome tool btw, thanks for all the work! :pray:

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -165,6 +165,24 @@ self: super:
   # Environment markers are not always included (depending on how a dep was defined)
   enum34 = if self.pythonAtLeast "3.4" then null else super.enum34;
 
+  eth-hash = super.eth-hash.overridePythonAttrs {
+    preConfigure = ''
+      substituteInPlace setup.py --replace \'setuptools-markdown\' ""
+    '';
+  };
+
+  eth-keyfile = super.eth-keyfile.overridePythonAttrs {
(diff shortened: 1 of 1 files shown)


adisbladis:
> I was a little confused why I needed to add the overrides for setuptools-markdown in e.g. eth-hash, because it's already present in the nixpkgs package definition, but maybe I'm just misunderstanding how poetry2nix does it's magic?

It does seem like you're misunderstanding things a bit :)
Poetry2nix is not using python packages from nixpkgs except as a _fallback_ mechanism, we're cleanly regenerating all packages in `poetry.lock`.

xwvvvvwx:
> Poetry2nix is not using python packages from nixpkgs except as a fallback mechanism for packages not in poetry.lock (like build-time only dependencies), we're cleanly regenerating all packages in poetry.lock.

Ah, right, that makes sense. Thanks! :sparkles:

Issue #154, opened 2020-08-19
Invalid version detection when using setuptools_scm + pyproject


cyraxjoe:
Apparently poetry2nix is unable to detect the proper version of the package when the package distributes a source package with `pyproject.toml` ([included by default ](https://setuptools.readthedocs.io/en/latest/history.html?#v43-0-0) in the `sdist` packages from setuptools)  and the package is using [setuptools_scm](https://github.com/pypa/setuptools_scm) with the `toml` extra, which relies on having the pyproject file in the source distribution.

An example of a package that will cause problems when used as a depedency in poetry2nix is [jaraco.functools](https://github.com/jaraco/jaraco.functools), which has [this pyproject.toml](https://github.com/jaraco/jaraco.functools/blob/master/pyproject.toml), building a wheel from the sdist of `jaraco.functools` either with `python setup.py wheel` or `pip wheel` _without_ the `pyproject.toml` will end up with a wheel with version `0.0.0`, messing up the following build, because the dependency at the expected version is not met.

The removal happens specifically at this hook script https://github.com/nix-community/poetry2nix/blob/master/hooks/pip-build-hook.sh#L8, I'm not sure what are the implications of _not_ removing the `pyproject.toml` but given the new default of including the pyproject in the source distribution, I think that more and more packages will start to relay in the presence of that file, the simple removal of the file seems that will cause additional _surprises_ in the future.

The issue can be replicated with:

Empty `fooapp.py` file.

`pyproject.toml`:
```toml
[tool.poetry]
name = "fooapp"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "^3.8"
"jaraco.functools" = "^3.0.1"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"

```
`poetry.lock`:
```toml
[[package]]
category = "main"
description = "Functools like those found in stdlib"
name = "jaraco.functools"
optional = false
python-versions = ">=3.6"
version = "3.0.1"

[package.dependencies]
more-itertools = "*"

[package.extras]
docs = ["sphinx", "jaraco.packaging (>=3.2)", "rst.linker (>=1.9)"]
[...]

PR #174, opened 2020-09-15
Add three overrides to make sure the pyproject.toml is not removed.

cyraxjoe:
Overrides: jaraco-functools, portend and tempora.

Closes issue #154


adisbladis:
Thanks!

PR #177, opened 2020-09-21
overrides.nix: handle zipp being null gracefully

akshaymankar:
If your poetry.lock file contains a marker making a dependency optional
on a specific python version, super.zipp is null, so getting its version
will fail. There's other code handling these being null (followups on
https://github.com/nix-community/poetry2nix/issues/50), so just make
this code handle it more gracefully, too.

Co-authored-by: Florian Klink <flokli@flokli.de>

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1133,25 +1133,25 @@ self: super:
     in
     if isWheel then wheelPackage else sourcePackage;
 
-  zipp =
-    (
-      if lib.versionAtLeast super.zipp.version "2.0.0" then (
-        super.zipp.overridePythonAttrs (
-          old: {
-            prePatch = ''
-              substituteInPlace setup.py --replace \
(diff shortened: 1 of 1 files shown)


flokli:
Hm, this now also fails on

```
Processing ./importlib_metadata-1.7.0-py2.py3-none-any.whl
ERROR: Could not find a version that satisfies the requirement zipp>=0.5 (from importlib-metadata==1.7.0) (from versions: none)
ERROR: No matching distribution found for zipp>=0.5 (from importlib-metadata==1.7.0)
```

I'm not sure why zipp isn't declared as a dependency for `importlib-metadata`.

adisbladis:
Thanks!

Side note: I wonder if there is a good way to generalise this? `zipp` is probably far from the only package we should apply a null check to.

PR #184, opened 2020-10-04
fix "invalid Git revision 'master'"

yihuang:
When running nix-shell after added a git dependency, it complains invalid Git revision 'master'.
I think it should always prefer `resolved_reference` to `reference`

### DIFF
--- a/bin/poetry2nix
+++ b/bin/poetry2nix
@@ -23,6 +23,11 @@ parser_lock.add_argument(
 
 
 def fetch_git(pkg):
+    try:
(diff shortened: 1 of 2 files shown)


adisbladis:
Thank you! I made a small fixup to make it work with older lock files by falling back to the old behaviour, but preferring `resolved_reference` if it exists.

yihuang:
can you make a release for this, I think the poetry version in nixpkgs master already generates `resolved_reference`.

PR #187, opened 2020-10-06
Add an override for torch

### LINKED ISSUE #69: torch fails in fixup phase
When building a poetry environment with just `torch` in it, it builds and installs just fine, but in the fixup phase (when it sets RPATHS), it fails to find CUDA-related libraries, specifically `libcuda.so.1`.

---
Also related, this still happens if I add an override to `export USE_CUDA=0` in the `preConfigure` phase, probably because it just downloads a wheel anyways?

Also also related, `pandas`, which is a dependency of `torch`, wouldn't build without cython present, so I had to override the `nativeBuildInputs` to include cython:
```nix
pandas = super.pandas.overrideAttrs (old: {
  nativeBuildInputs = old.nativeBuildInputs ++ [ self.cython ];
});
```
---

Here are my [`pyproject.toml`](http://ix.io/2cWS) and [`poetry.lock`](http://ix.io/2cWT) files, and here is the [output of `nix-shell`](http://ix.io/2cWU).

If I add the below override to export `LD_LIBRARY_PATH` so it can find those libraries, I get one that I can't locate (`libcudart.so.10.1`).

```nix
(self: super: {
  torch = super.torch.overrideAttrs (old: {
    nativeBuildInputs = old.nativeBuildInputs
     ++ [ pkgs.cudatoolkit pkgs.linuxPackages.nvidia_x11 ];
    preConfigure = ''
      export LD_LIBRARY_PATH="${pkgs.linuxPackages.nvidia_x11}/lib:${pkgs.cudatoolkit}/targets/x86_64-linux/lib"
    '';
})
```
Output from running `nix-shell` with that is [here](http://ix.io/2cWX).

I'm not really sure where to go from here, or even if I'm going in the right direction.
I don't mind dropping the CUDA support when building pytorch, but like I said above exporting the proper variable doesn't affect it.

timokau:
Mostly aggregating the suggestions from
https://github.com/nix-community/poetry2nix/issues/69.

Fixes #69.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1115,6 +1115,36 @@ self: super:
     }
   );
 
+  torch = lib.makeOverridable
+    ({ enableCuda ? false
+     , cudatoolkit ? pkgs.cudatoolkit_10_1
+     , pkg ? super.torch
+     }: pkg.overrideAttrs (old:
+      {
+        preConfigure =
+          if (!enableCuda) then ''
+            export USE_CUDA=0
+          '' else ''
+            export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${cudatoolkit}/targets/x86_64-linux/lib"
+          '';
+        preFixup = lib.optionalString (!enableCuda) ''
+          # For some reason pytorch retains a reference to libcuda even if it
+          # is explicitly disabled with USE_CUDA=0.
+          find $out -name "*.so" -exec ${pkgs.patchelf}/bin/patchelf --remove-needed libcuda.so.1 {} \;
+        '';
+        buildInputs = old.buildInputs ++ lib.optionals enableCuda [
+          pkgs.linuxPackages.nvidia_x11
+          pkgs.nccl.dev
+          pkgs.nccl.out
+        ];
+        propagatedBuildInputs = [
+          super.numpy
+          super.future
+        ];
+      })
+    )
+    { };
+
   # nix uses a dash, poetry uses an underscore
   typing_extensions = super.typing_extensions or self.typing-extensions;
 


adisbladis:
Could you make cuda support optional?

timokau:
I managed to make it work without cuda. For now the configuration is hardcoded though, so it should not be merged as-is. Is there some standard configuration mechanism for `poetry2nix`?

overrides.nix
```diff
@@ -1115,6 +1115,34 @@ self: super:
     }
   );
 
+  torch = super.torch.overrideAttrs (old:
```
Mic92, line 1118: You could use `lib.makeOverridable`, that way users could provide their own override that overrides this flags in the `overrides`.

timokau: Like this?

Mic92: Maybe? Does it work for you?

overrides.nix
```diff
@@ -1115,6 +1115,36 @@ self: super:
...
+          find $out -name "*.so" -exec ${pkgs.patchelf}/bin/patchelf --remove-needed libcuda.so.1 {} \;
+        '';
+        buildInputs = old.buildInputs ++ lib.optionals enableCuda [
+          pkgs.linuxPackages.nvidia_x11
+          pkgs.nccl.dev
+          pkgs.nccl.out
+        ];
+        propagatedBuildInputs = [
+          super.numpy
+          super.future
```
Mic92, line 1142: poetry2nix is not able to figure out this part correctly?

overrides.nix
timokau: Yes, it works for me. At least the default setting without cuda. I don't have the setup to test the CUDA support.

overrides.nix
timokau: No, for some reason it is not. The build fails when I'm not explicitly specifying this.

overrides.nix
Mic92: I feel a bit uneasy to have untested code. Could you maybe leave a TODO which points to the commit where you introduce CUDA? Than someone who can actually test built it with CUDA can pick it up.

Mic92: Or do you mean you don't have a graphics card to test it?

timokau: Yes, that is what I mean. It compiles fine. I just noticed an issue where the cuda/no-cuda logic was inverted though, that is fixed now. I just cannot test the cuda version at runtime.

Issue #197, opened 2020-10-26
mkPoetryDeps treats git branch as hash


lukebfox:
Presumably due to recent lockfile format changes, attempting to re-enter nix-shell after relocking a project with git dependencies fails.
Steps to reproduce:
```bash
$ git clone https://github.com/NixOS/nixops-aws
$ cd nixops-aws
$ nix-shell 
$ poetry lock
$ exit
$ nix-shell
```
For this repository the pyproject entry causing the issue is nixops,
```toml
[tool.poetry.dependencies]
nixops = {git="https://github.com/nixos/nixops" rev="master"}
```
and attempting to enter nix-shell after relocking results in the following error:
```
error: --- BadHash ------------------------------------------------------------------ nix-shell
hash 'master' has wrong length for hash type 'sha1'
(use '--show-trace' to show detailed location information)
```
Similarly when I try another variant,
```toml
nixops = {git="https://github.com/lukebfox/nixops" rev="statedict-store-dict"}
```
I get another error:
```
error: --- UsageError --------------------------------------------------------------- nix-shell
unknown hash algorithm 'statedict'
Try 'nix-shell --help' for more information.
```
The top of the stack trace in both cases:
```
trace: while evaluating the attribute 'src.name'
at: (155:7) in file: /nix/store/vjj6vlbfq661l2zllgvyspdsdyvz8hl9-source/pkgs/development/tools/poetry2nix/poetry2nix/mk-poetry-dep.nix

   154|       # Here we can then choose a file based on that info.
   155|       src =
      |       ^
   156|         if isGit then

trace: while evaluating 'hasSuffix'
at: (212:5) in file: /nix/store/vjj6vlbfq661l2zllgvyspdsdyvz8hl9-source/lib/strings.nix

   211|     # Input string
   212|     content:
      |     ^
   213|     let

trace: from call site
at: (122:25) in file: /nix/store/vjj6vlbfq661l2zllgvyspdsdyvz8hl9-source/pkgs/development/interpreters/python/mk-python-derivation.nix

   121|       pythonRemoveBinBytecodeHook
   122|     ] ++ lib.optionals (lib.hasSuffix "zip" (attrs.src.name or "")) [
      |                         ^
   123|       unzip
```
(it goes on a fair bit)

Not sure if this is intended behaviour and the pyproject.toml needs updating to something new, or this is an issue.

teto:
I've just noticed the same issue with `tag` instead of `rev`: it's used as a hash:
```
[...]

PR #199, opened 2020-10-28
fix when fetching tag and unspecified branch

teto:
follow up of https://github.com/nix-community/poetry2nix/pull/167 . Explanation at https://github.com/nix-community/poetry2nix/issues/197 . I would like to update the tests but not sure what's best.

### DIFF
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -155,27 +155,22 @@ pythonPackages.callPackage
       # We need to retrieve kind from the interpreter and the filename of the package
       # Interpreters should declare what wheel types they're compatible with (python type + ABI)
       # Here we can then choose a file based on that info.
-      src =
+      src = let
+          ref = if (sourceSpec ? tag)
+                then "refs/tags/${sourceSpec.tag}"
+                else (sourceSpec.branch or null);
+        in
         if isGit then
           (
-            builtins.fetchGit {
+            builtins.fetchGit (lib.debug.traceValSeq ({
               inherit (source) url;
               rev = source.resolved_reference or source.reference;
-              ref = sourceSpec.branch or sourceSpec.rev or sourceSpec.tag or "HEAD";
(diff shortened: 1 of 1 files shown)


teto:
It doesn't seem it's possible in poetry to specify both a rev and a ref ? for htis reason I can't fetch some specific revisions that are not on the default branch.

mk-poetry-dep.nix
```diff
@@ -152,14 +152,18 @@ pythonPackages.callPackage
...
       # Interpreters should declare what wheel types they're compatible with (python type + ABI)
       # Here we can then choose a file based on that info.
-      src =
+      src = let
+          # tagRef = if (sourceSpec ? tag) then "refs/tags/${sourceSpec.tag}" else null;
+        in
         if isGit then
           (
-            builtins.fetchGit {
+            builtins.fetchGit (lib.debug.traceValSeq ({
```
Mic92, line 160: This should be not left it no, no?

teto: yes it should go away :)  I am not sure this whole PR is a definitive fix nor how to operate/update the tests so I dont intend to complete this as of now. Anyone feel free to pick it up.

teto:
I think I have my final solution, if and only if there is a branch or tag add { ref = REF } (with REF=refs/tags/TAG if a tag) and nothing otherwise (I tried to pass ref=null instead of ref="HEAD": but nix doesnt like it).
I would like to see a fix merged but I don't trust much the test infra at the moment, should I ?

adisbladis:
> I think I have my final solution, if and only if there is a branch or tag add { ref = REF } (with REF=refs/tags/TAG if a tag) and nothing otherwise (I tried to pass ref=null instead of ref="HEAD": but nix doesnt like it).

This has now been merged on master in https://github.com/nix-community/poetry2nix/pull/240

Issue #201, opened 2020-11-03
Bump poetry version to 1.1.4


takeda:
The version 1.1.0 is quite buggy, since it introduced new features. I personally run into this bug: https://github.com/stan-dev/httpstan/pull/466

Can we bump the poetry version to latest one? I'm not sure how I can properly override it since it uses pyproject.toml file instead of having a standard derivation.

PR #202, opened 2020-11-03
Bumped poetry to version 1.1.4

takeda:
Fixes https://github.com/nix-community/poetry2nix/issues/201

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -17,5 +17,6 @@ pkgs.mkShell {
     pkgs.nixpkgs-fmt
     pkgs.poetry
     pkgs.niv
+    pkgs.jq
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -17,5 +17,6 @@ pkgs.mkShell {
     pkgs.nixpkgs-fmt
     pkgs.poetry
     pkgs.niv
+    pkgs.jq
```
Mic92, line 20: Short question, where is `jq` used here?

takeda: sorry, I just noticed this, it is used by the update script `pkgs/poetry/update`

PR #205, opened 2020-11-05
Support source url type

Mic92:
depends on https://github.com/nix-community/poetry2nix/pull/204

### DIFF
--- a/bin/poetry2nix
+++ b/bin/poetry2nix
@@ -7,35 +7,84 @@ import toml
 import json
 import sys
 
-from typing import Dict, Any, Tuple
+from typing import Dict, Any, Tuple, List
 
 
-def fetch_git(pkg: Dict[str, Any]) -> Tuple[str, subprocess.CompletedProcess]:
-    try:
-        reference = pkg["source"]["resolved_reference"]
-    except KeyError:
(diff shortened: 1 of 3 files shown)


Mic92:
I think I need to switch to fetchzip here.

NorfairKing:
Hi @Mic92. I tried using a url-based package. Is it possible that the `bulitins.fetchTarball` on line 158 of `mk-poetry-dep.nix` should be a `fetchzip` instead? I'm getting a strange error when pointing the url at a `.whl` file.

This is in my `pyproject.toml`:
```
torchvision = { url = "https://download.pytorch.org/whl/cu111/torchvision-0.9.1%2Bcu111-cp39-cp39-linux_x86_64.whl" }
```


This is the error:

```
nix-shell
unpacking 'https://download.pytorch.org/whl/cu111/torchvision-0.9.1%2Bcu111-cp39-cp39-linux_x86_64.whl'...
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
error: program 'tar' failed with exit code 2
```

CC @nh2

NorfairKing:
Scratch that, I tried that out and ran into different problems, see #301

PR #210, opened 2020-11-22
Fix flake

typetetris:
At the moment

```
nix flake show --no-write-lock-file github:nix-community/poetry2nix
```

fails with

```
* Added 'nixpkgs': 'github:NixOS/nixpkgs/e286f0cf3bcffd5c0f006bb1831563d63b281fb2'
github:nix-community/poetry2nix/1894b501cf4431fb218c4939a9acdbf397ac1803
├───overlay: Nixpkgs overlay
└───packages
    ├───aarch64-linux
    │   ├───poetry: package 'python3.8-poetry-1.1.4'
error: --- Error ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
expected a derivation
```

So it seams stuff in `packages` output need to be a derivation.

To turn poetry2nix into a proper flake, just remove the packages. Its not needed anyways and just forces people to download nixpkgs  checkout another time.

Of course flakes is experimental and such sort of breakage is to be expected, I think it would be good to fix it up anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,32 @@ in pkgs.poetry2nix.mkPoetryApplication {
 }
 ```
 
+## Using the flake
+
+For the experimental flakes functionality we provide poetry2nix as a flake providing an overlay
+to use with [nixpkgs](https://nixos.org/nixpkgs/manual).
+
+#### Example
+
+```nix
+{
+    description = "Your flake using poetry2nix";
+
+    inputs.nixpkgs.url = "github:NixOS/nixpkgs";
+    inputs.utils.url = "github:numtide/flake-utils";
+    inputs.poetry2nix-src.url = "github:nix-community/poetry2nix";
+
+    outputs = {nixpkgs, utils, poetry2nix-src, self}: utils.lib.eachDefaultSystem (system: let
+
+      pkgs = import nixpkgs { inherit system; overlays = [ poetry2nix-src.overlay ]; };
+
+    in
+      {
+         # use pkgs.poetry2nix here.
+      });
+  }
+```
+
 ## Contributing
 
 Contributions to this project are welcome in the form of GitHub PRs. Please consider the following before creating PRs:
--- a/flake.nix
+++ b/flake.nix
@@ -1,30 +1,8 @@
 {
   description = "Poetry2nix flake";
 
-  outputs = { self, nixpkgs }:
-    let
-      # TODO: There must be a better way to provide arch-agnostic flakes..
-      systems = [ "x86_64-linux" "i686-linux" "x86_64-darwin" "aarch64-linux" ];
-      forAllSystems = f: nixpkgs.lib.genAttrs systems (system: f system);
-      # Memoize nixpkgs for different platforms for efficiency.
-      nixpkgsFor = forAllSystems (
-        system:
-        import nixpkgs {
-          inherit system;
-          overlays = [ self.overlay ];
-        }
-      );
-    in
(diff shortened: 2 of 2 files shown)


teto:
isn't it more the problem of the incorrect system ? I like to use flakes to run the latest version nix build .# or `nix run `. Is it fine if I revert this and fixes the flake instead ?

teto:
I went ahead and did it, you just had to remove poetry2nix from the list of packages :s

typetetris:
Isn't `poetry` just passed through from the `nixpkgs` input now?

If that's the case, users now have to use

`inputs.poetry2nix.inputs.nixpkgs.follows = "nixpkgs";`

to not download two different versions of nixpkgs.

If only the overlay is provided, `nixpkgs` as input is unnecessary at all. Which I thought, was a nice thing.

teto:
with only the overlay, you don't exploit the flake ability to be able to run poetry from master. With the linked PR, you can run 
 `nix run github:nix-community/poetry2nix` to run poetry from master etc. To not redownload nixpkgs, you can override the inputs or lock poetry to download it just once (or do "follow" as you mentioned).

typetetris:
Wouldn't it be more appropriate to have poetry from master as a separate flake then?

teto:
no, if you want to run a specific version, you can lock the flake to a specific revision already.

typetetris:
I can lock which flake? Is there a flake for poetry only (without poetry2nix) already?

teto:
I feel like we dont understand eachother ^^'' . What are you trying to do ? poetry is already available in nixpkgs (try `nix build nixpkgs#poetry`). If you use this repo's flake, it means you want a newer version. But if you want a version in-between you can still pin the flake as shown in https://nixos.wiki/wiki/Flakes (search for `# Pin flakes to a specific revision`).

typetetris:
I think, maybe, I am just a bit stubborn here.

> I feel like we dont understand eachother ^^'' .

I have the impression too.

> What are you trying to do ?

I want to be able to use poetry2nix, the set of nix functions, with minimal hassle.

And I don't always need poetry.

I need poetry only while developing a new package, but then I can easily get it by other means.

If it is necessary to have nixpkgs as an additional input in the flake to provide a more recent version of poetry in its output, I think it is best to skip that.

Who wants a more recent version of poetry can either use a flake containing only poetry or use the overlay and access the
package that way.

> poetry is already available in nixpkgs (try `nix build nixpkgs#poetry`).

At a certain version yes.

> If you use this repo's flake, it means you want a newer version.

No. I want to use the "nix library" providing the `poetry2nix` set of functions.

> But if you want a version in-between you can still pin the flake as shown in https://nixos.wiki/wiki/Flakes (search for `# Pin flakes to a specific revision`).

That was never my concern.

Reading through all that, I think the real decision is,

1. wether a user of the poetr2ynix flake should always have to write

    inputs.poetry2nix.input.nixpkgs.follow = "nixpkgs";

2. or only in the cases, where he wants to have a more recent version of poetry:

      pkgs = import nixpkgs { inherit system; overlays = [ poetry2nix-src.overlay ]; };

I thought number 2. better as it only effects people, who want to use the more recent version of poetry and is a well known syntax, while 1. has to be done by everybody, can be easily forgot.

teto:
> No. I want to use the "nix library" providing the poetry2nix set of functions.

Isn't the repo synced with nixpkgs ? mkPoetryApplication and all are available from nixpkgs ?

PR #211, opened 2020-11-23
overrides: add colour

rissson:


overrides.nix
```diff
@@ -106,6 +106,12 @@ self: super:
     }
   );
 
+  colour = super.colour.overridePythonAttrs (
+    old: {
+      buildInputs = old.buildInputs ++ [ self.2dto1 ];
```
adisbladis, line 111:
```suggestion
      buildInputs = old.buildInputs ++ [ self."2dto1" ];
```

adisbladis (changes requested):

overrides.nix
rissson: I'm curious as why this was requested though

adisbladis: Because Nix attrnames cannot start with a number except when quoted.
But it turns out this fix was _also_ wrong as you had an initial typo, the correct spelling should be `d2to1`.

rissson: Yes indeed, I must be really bad at copy-pasting. Fixed it!

PR #213, opened 2020-11-24
fix cross builds

LnL7:
Make poetry2nix builds work when cross compiling.


LnL7:
Looks like github actions needs to be bumped.

PR #214, opened 2020-11-25
fix(flake): overlay args are standardized

teto:
when trying to use poetry2nix overlay in my flake, I get:
error: --- Error --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
overlay does not take an argument named 'final'
------------------------------------------------------------------------------------------ show-trace ------------------------------------------------------------------------------------------
trace: while checking the overlay 'overlays.poetry2nix'

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -1,7 +1,7 @@
-self: super: {
+final: prev: {
 
-  poetry2nix = import ./default.nix { pkgs = self; poetry = self.poetry; };
+  poetry2nix = import ./default.nix { pkgs = final; poetry = final.poetry; };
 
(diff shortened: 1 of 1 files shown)


adisbladis:
I hate that flakes enforces these names for no good reason.... But whatever.. I'll merge it.

adisbladis:
Merged in 1a6885076a33a3065b3aa0229e30a65ce69c4e30.

teto:
I kinda like it; makes copy/pasting easier: standardization is a lot of what flakes are about even though this case is not the best example. Thanks for the merge

PR #221, opened 2020-12-01
Flake default app

teto:
reverts https://github.com/nix-community/poetry2nix/pull/210 and make it possible to use the flake via `nix run`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1609246779,
-        "narHash": "sha256-eq6ZXE/VWo3EMC65jmIT6H/rrUc9UWOWVujkzav025k=",
+        "lastModified": 1610051610,
+        "narHash": "sha256-U9rPz/usA1/Aohhk7Cmc2gBrEEKRzcW4nwPWMPwja4Y=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "08c7ad4a0844adc4a7f9f5bb3beae482e789afa4",
+        "rev": "3982c9903e93927c2164caa727cd3f6a0e6d14cc",
         "type": "github"
       },
       "original": {
@@ -17,11 +17,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1609542771,
-        "narHash": "sha256-OcV1R1wnVDHFOzBib8tbPAbfDU5D3G6yIXV4Kr036XI=",
+        "lastModified": 1610729867,
+        "narHash": "sha256-bk/SBaBLqZX/PEqal27DMQwAHHl0dcZMp8NNksQr80s=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "014440d7105ce2be945b79dc957dd08720a9c029",
+        "rev": "04af07c659c6723a2259bb6bc00a47ec53330f20",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,42 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1609246779,
+        "narHash": "sha256-eq6ZXE/VWo3EMC65jmIT6H/rrUc9UWOWVujkzav025k=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


teto:
@Mic92  do you have merge rights ? or anyone else than @adisbladis ? seems like his bandwith is full xD

flake.nix
```diff
@@ -1,8 +1,37 @@
...
+      forAllSystems = f: nixpkgs.lib.genAttrs systems (system: f system);
+      # Memoize nixpkgs for different platforms for efficiency.
+      nixpkgsFor = forAllSystems (
+        system:
+        import nixpkgs {
+          inherit system;
+          overlays = [ self.overlay ];
+        }
+      );
+    in
```
Mic92, line 17: You can use flake-utils here: https://github.com/numtide/flake-utils

Mic92: This would be the preferred way since it also comes with a `defaultSystems` implementation that would not require an update from `poetry2nix` side.

adisbladis:
@Mic92 Feel free to merge this when you think it's good. I don't know enough about flakes to review this right now.

Mic92:
I would like to see https://github.com/nix-community/poetry2nix/pull/221#discussion_r545822532 and than it should be good to go.

flake.nix
```diff
@@ -1,8 +1,37 @@
...
+
+      packages = forAllSystems (system: {
+        inherit (nixpkgsFor.${system}) poetry;
+      }
+      );
+
+      apps = forAllSystems (system: {
+        poetry = {
+          type = "app";
+          program = self.packages."${system}".poetry + "/bin/poetry";
```
Mic92, line 29: Why is poetry the default app here instead of `poetry2nix`?

teto: I never had to use poetry2nix, more like the nix code (mkPoetryEnv etc). This could be added afterwards

teto:
I rebased with flake-utils. The CI error looks unrelated ?

Mic92:
Can you also update flake.lock?

```
$ nix run 'github:teto/poetry2nix/flake-default-app'
error: --- Error ---------------------------------------------------------------------------------------------------------
cannot write modified lock file of flake 'github:teto/poetry2nix/flake-default-app' (use '--no-write-lock-file' to ignore)
```

teto:
There is no .lock file on master, I think @adisbladis was against it, should I still add one ?

Mic92:
> There is no .lock file on master, I think @adisbladis was against it, should I still add one ?

Without .lock file the flake is not usable.

teto:
I added the lock file, tested `nix build` / `nix run` / `nix flake check`. I would prefer to postpone adding poetry2nix to the list of packages in another PR.

flake.nix
```diff
@@ -1,8 +1,28 @@
 {
   description = "Poetry2nix flake";
 
-  outputs = { self }:
+  inputs.flake-utils.url = "github:numtide/flake-utils";
```
Mic92, line 4:
```suggestion
  inputs.nixpkgs.url = "github:NixOS/nixpkgs";
  inputs.flake-utils.url = "github:numtide/flake-utils";
```

flake.lock
```diff
@@ -0,0 +1,38 @@
...
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "narHash": "sha256-WeuJZ0I5kjSa+xxxEMFNvBdlZCwktEKZBNexwWe6718=",
+        "path": "/home/teto/nixpkgs",
```
Mic92, line 21: This is not correct. please re-run `nix flake update --recreate-lock-file` after adding the line below.

teto: oops sry. Fixed

teto:
failure unrelated ?

flake.nix
```diff
@@ -1,8 +1,29 @@
...
+        myNixpkgs = import nixpkgs {
+          inherit system;
+          overlays = [ self.overlay ];
+        };
+      in
+      rec {
+        packages = {
+          inherit (myNixpkgs) poetry;
+        };
+        defaultPackage = packages.poetry;
```
Mic92, line 21: Why is the default app poetry and not poetry2nix?

Mic92: poetry is already available in nixpkgs itself.

teto: this repo provides a newer one generally. I haven't had the need to run poetry2nix yet. feel free to change it in a follow up PR.

Mic92: solved.

Issue #186, opened 2020-10-05
Add doc on how to create a shell that includes dev-dependencies 


mickours:
It is possible to create a nix-shell that includes all the development dependencies but I can't find any example.

The use case is: I want the CI to lint/format/test the code with de dependencies declared as in the `pyproject.toml` file as `dev-dependencies`.

The doc suggest that this is the case with `mkPoetryEnv`:

> with all dependencies declared by the designated poetry project and lock files

I finally found the trick:
```nix
let
    envShell = poetry2nix.mkPoetryEnv {
      projectDir = ./.;
    };
in envShell.env
```

Related to #60

mickours:
Also, adding additional non-python dependencies in the shell can be done with:
```nix
let
    envShell = poetry2nix.mkPoetryEnv {
      projectDir = ./.;
    };
  in pkgs.mkShell {
    buildInputs = [ envShell pkgs.hello ];
  }
```

Issue #60, opened 2020-02-26
Write Usage documentation


gilligan:
The README now has an API section, the next step is to add a Usage segment which should also cover overrides, the different ways to override (with defaults, without, with modified overrides, without any overrides) and mention the possibility to use `preferWheel`

jtrakk:
I've also been wondering how to install it.

surajbarkale:
Here is a shell.nix example https://github.com/mjlbach/nix-shells/blob/master/lorri_tensorflow_nixgl_poetry2nix/shell.nix

PR #225, opened 2020-12-13
Add mkPoetryEnv shell.nix examples to README

kylegentle:
It took me a while to figure this out, so I think it may be a useful reference for future `poetry2nix` users.

Related to and inspired by #186, #60.


adisbladis:
This is excellent, thank you!

PR #235, opened 2021-01-15
gitignore: ignore direnv

cpcloud:
This PR adds `.direnv` to the repository `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 result*
 artifacts
 poetry2nix-flamegraph.svg
+.direnv


adisbladis:
What is creating this file? Why should we ignore it?

cpcloud:
Ah, sorry I should've given a bit more context. This file is created by the direnv tool, kicked off by running `direnv allow` in a directory with an `.envrc` such as a clone of this repo.

It should be ignored for a variety of reasons, but I think probably the one that sticks out to me the most is that its not meant to be shared with anyone.

adisbladis:
This file is _not_ created by direnv but by https://github.com/nix-community/nix-direnv/.
Regardless, this change seems sensible.

Could you add a reference to https://github.com/nix-community/nix-direnv/ in your commit message?

cpcloud:
@adisbladis Thanks! Will do.

cpcloud:
@adisbladis Done! Thanks again for the review.

PR #239, opened 2021-01-22
[DNM] remove wheel from pyproject.toml in tests

mjlbach:
All tests now succeed pending #240 

I removed wheel from pyproject.toml, and removed boto due to an upstream issue https://github.com/python-poetry/poetry/issues/3367


adisbladis:
This was caused by a bug in the `wheel` override and has been properly fixed in 261ea2b15a3a281d204b90d0c3e36e1d621a4f88.

PR #242, opened 2021-01-26
Fix stdenv.platform accesses for latest nixpkgs

K900:
This fixes builds being completely broken after:
* https://github.com/NixOS/nixpkgs/pull/110726
* https://github.com/NixOS/nixpkgs/pull/110544

I'm not sure if this should be handled in a backwards compatible way, but if it is, I'd be happy to adjust the patch.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -1,4 +1,4 @@
-{ lib, pkgs }:
+{ lib, pkgs, stdenv }:
 let
   inherit (import ./semver.nix { inherit lib ireplace; }) satisfiesSemver;
   inherit (builtins) genList length;
@@ -194,6 +194,23 @@ let
         inherit src;
       };
     };
+
+  # Maps Nixpkgs CPU values to target machines known to be supported for manylinux* wheels.
+  # (a.k.a. `uname -m` output from CentOS 7)
+  #
+  # This is current as of manylinux2014 (PEP-0599), and is a superset of manylinux2010 / manylinux1.
+  # s390x is not supported in Nixpkgs, so we don't map it.
+  manyLinuxTargetMachines = {
+    x86_64 = "x86_64";
+    i686 = "i686";
+    aarch64 = "aarch64";
+    armv7l = "armv7l";
+    powerpc64 = "ppc64";
+    powerpc64le = "ppc64le";
+  };
+
+  # Machine tag for our target platform (if available)
+  targetMachine = manyLinuxTargetMachines.${stdenv.targetPlatform.parsed.cpu.name} or null;
 in
 {
   inherit
@@ -207,5 +224,6 @@ in
     cleanPythonSources
     moduleName
     getPythonVersion
+    targetMachine
     ;
 }
--- a/pep425.nix
+++ b/pep425.nix
@@ -1,6 +1,7 @@
-{ lib, stdenv, python, isLinux ? stdenv.isLinux }:
+{ lib, stdenv, poetryLib, python, isLinux ? stdenv.isLinux }:
 let
   inherit (lib.strings) hasSuffix hasInfix splitString removeSuffix;
+  inherit (poetryLib) targetMachine;
 
   # The 'cpxy" as determined by `python.version`
   #
@@ -72,12 +73,16 @@ let
       withPlatform =
         if isLinux
         then
-          (
-            x: x.platform == "manylinux1_${stdenv.platform.kernelArch}"
-              || x.platform == "manylinux2010_${stdenv.platform.kernelArch}"
-              || x.platform == "manylinux2014_${stdenv.platform.kernelArch}"
-              || x.platform == "any"
-          )
+          if targetMachine != null
+          then
+            (
+              x: x.platform == "manylinux1_${targetMachine}"
+                || x.platform == "manylinux2010_${targetMachine}"
+                || x.platform == "manylinux2014_${targetMachine}"
+                || x.platform == "any"
+            )
+          else
+            (x: x.platform == "any")
         else (x: hasInfix "macosx" x.platform || x.platform == "any");
       filterWheel = x:
         let
--- a/pep508.nix
+++ b/pep508.nix
@@ -1,6 +1,6 @@
 { lib, stdenv, poetryLib }: python:
 let
-  inherit (poetryLib) ireplace;
+  inherit (poetryLib) ireplace targetMachine;
 
   # Like builtins.substring but with stop being offset instead of length
   substr = start: stop: s: builtins.substring start (stop - start) s;
@@ -95,7 +95,7 @@ let
           else if stdenv.isDarwin then "darwin"
           else throw "Unsupported platform"
         );
-        platform_machine = stdenv.platform.kernelArch;
+        platform_machine = targetMachine;
         platform_python_implementation =
           let
             impl = python.passthru.implementation;
--- a/default.nix
+++ b/default.nix
@@ -1,7 +1,7 @@
 { pkgs ? import <nixpkgs> { }
 , lib ? pkgs.lib
 , poetry ? null
-, poetryLib ? import ./lib.nix { inherit lib pkgs; }
+, poetryLib ? import ./lib.nix { inherit lib pkgs; stdenv = pkgs.stdenv; }
 }:
 let
   inherit (poetryLib) isCompatible readTOML moduleName;
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -31,7 +31,7 @@ pythonPackages.callPackage
       inherit (poetryLib) isCompatible getManyLinuxDeps fetchFromPypi moduleName;
 
       inherit (import ./pep425.nix {
-        inherit lib python;
+        inherit lib poetryLib python;
         inherit (pkgs) stdenv;
       }) selectWheel
         ;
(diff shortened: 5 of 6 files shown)


K900:
The tests are currently failing because they're presumably running on an older Nixpkgs commit.

K900:
I'm not sure why the tests are failing now, but master seems to have the same issue.

cpcloud:
@adisbladis Any chance of getting this in any time soon?

pep425.nix
```diff
@@ -73,9 +74,9 @@ let
         if isLinux
         then
           (
-            x: x.platform == "manylinux1_${stdenv.platform.kernelArch}"
-              || x.platform == "manylinux2010_${stdenv.platform.kernelArch}"
-              || x.platform == "manylinux2014_${stdenv.platform.kernelArch}"
+            x: x.platform == "manylinux1_${targetArch}"
+              || x.platform == "manylinux2010_${targetArch}"
+              || x.platform == "manylinux2014_${targetArch}"
```
Infinisil, line 79: We can just use `stdenv.hostPlatform.parsed.cpu.arch` instead, no need to introduce this into `poetryLib` then, and that's compatible with both stable and unstable.

Infinisil (changes requested):

pep425.nix
K900: I don't think this works: for example, x86_64 has `cpu.arch` set to [`x86-64`](https://github.com/NixOS/nixpkgs/blob/8baac2af752450b20cb8b37520059ecd86490b8b/lib/systems/parse.nix#L89), whereas we specifically want `x86_64`. Now that I'm looking at it, I'm pretty sure `linuxArch` doesn't work either - for example, it uses [`arm`](https://github.com/NixOS/nixpkgs/blob/8baac2af752450b20cb8b37520059ecd86490b8b/lib/systems/default.nix#L90) for aarch32, but PyPI uses `armv7l`. I guess the most correct solution would be to introduce a custom mapping from `stdenv.hostPlatform.parsed.cpu` to PyPI tags...

Infinisil: Ah I think we can use `stdenv.hostPlatform.parsed.cpu.name` then

K900: Not quite - `powerpc64` is `ppc64` on PyPI :(

Infinisil: Where do you see that?

K900: Python ABI tags use the exact architecture specifiers `uname` uses, so Linux's `ARCH`. For manylinux2014, they're listed here: https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy

K900: More specifically, we're looking at

```
x86_64
i686
aarch64
armv7l
ppc64
ppc64le
s390x
```

K900: Nixpkgs doesn't run on s390x AFAIK, so we only need to check for 6 platforms.

K900: (also, manylinux1 and manylinux2010 are restricted to x86_64 and i686 only, but we can still check for every architecture, as manylinux{1,2010} wheels for other architectures can't be uploaded to PyPI)

Infinisil: I think using `parsed.cpu.name` should be good enough for now. It's correct for the most important ones, and notably also for more than the previous `stdenv.platform.kernelArch`. On 20.09:
```
❯ ne -A stdenv.platform.kernelArch --argstr system x86_64-linux
"x86_64"
❯ ne -A stdenv.platform.kernelArch --argstr system i686-linux
"i386"
❯ ne -A stdenv.platform.kernelArch --argstr system aarch64-linux
"arm64"
❯ ne -A stdenv.platform.kernelArch --argstr system armv7l-linux
"arm"
```

But for `stdenv.hostPlatform.parsed.cpu.name`:
```
❯ ne -A stdenv.hostPlatform.parsed.cpu.name --argstr system x86_64-linux
"x86_64"
❯ ne -A stdenv.hostPlatform.parsed.cpu.name --argstr system i686-linux
"i686"
❯ ne -A stdenv.hostPlatform.parsed.cpu.name --argstr system aarch64-linux
"aarch64"
❯ ne -A stdenv.hostPlatform.parsed.cpu.name --argstr system armv7l-linux
"armv7l"
```

K900: This still gets us `powerpc` and `powerpc64` though. I'd prefer to add an explicit lookup table, especially since it's going to be very small. Is there something I'm missing that makes it a bad idea?

Infinisil: I think these values should ideally come from nixpkgs, to keep platform handling out of poetry2nix. From the PEP's, it looks like these values come from CentOS (for PEP425), or from `platform.machine() -> uname -m` (for PEP508). Nixpkgs notably already has a partial `uname` mapping. These all don't seem python specific.

How about we just use `parsed.cpu.name` for now to keep it simple, but you can make a nixpkgs PR for adding such mappings (where you can also ping the people who know more about platform stuff), and once it's been in nixpkgs for a while we can switch poetry2nix to use that.

K900: > I think these values should ideally come from nixpkgs, to keep platform handling out of poetry2nix.

Agreed conceptually, but I'd rather have a mapping in place now and then switch to nixpkgs' one when it's good enough. Otherwise we're basically knowingly breaking PowerPC on all stable NixOS versions and then however long it takes for nixpkgs to get a proper mapping.

> From the PEP's, it looks like these values come from CentOS (for PEP425), or from platform.machine() -> uname -m (for PEP508)

The CentOS values are also just `uname -m` from CentOS machines, so this shouldn't be a problem.

> Nixpkgs notably already has a partial uname mapping. These all don't seem python specific.

I was actually looking for a `uname -m` mapping in nixpkgs and couldn't find one. Do you have a link to the one you mentioned? (I've seen [this](https://github.com/NixOS/nixpkgs/blob/8baac2af752450b20cb8b37520059ecd86490b8b/lib/systems/default.nix#L73), but that's not helpful unfortunately).

> How about we just use parsed.cpu.name for now to keep it simple, but you can make a nixpkgs PR for adding such mappings (where you can also ping the people who know more about platform stuff), and once it's been in nixpkgs for a while we can switch poetry2nix to use that.

I'm up to do that (and discuss it on the nixpkgs side), but again, I'd rather have a temporary mapping in place in poetry2nix for now, or at the very least, some sort of a comment that explicitly mentions the PowerPC issue.

Infinisil: Regarding the `uname` mapping, I said "partial" because it notably doesn't include all the `uname` flags, such as `-m` :)

Considering that I don't believe anybody has even tested poetry2nix with powerpc64, I think it wouldn't be a problem to just not support it for now until we have the correct mapping in nixpkgs. So I'd just go for an assertion that checks the cpu name to be one of the supported ones.

But if you'd much rather have a mapping directly in poetry2nix, that's fine too I guess. Feel free to go for it.

K900: All right, I'll try to write up a patch in a few hours. Thanks!

K900: ...actually took me a lot less than a few hours.

lib.nix
```diff
@@ -194,6 +194,23 @@ let
...
+    x86_64 = "x86_64";
+    i686 = "i686";
+    aarch64 = "aarch64";
+    armv7l = "armv7l";
+    powerpc64 = "ppc64";
+    powerpc64le = "ppc64le";
+  };
+
+  # Machine tag for our target platform (if available)
+  targetMachine = manyLinuxTargetMachines.${stdenv.targetPlatform.parsed.cpu.name} or null;
```
Infinisil, line 213: Instead of `null`, an error should be thrown

pep508.nix
```diff
@@ -95,7 +95,7 @@ let
           else if stdenv.isDarwin then "darwin"
           else throw "Unsupported platform"
         );
-        platform_machine = stdenv.platform.kernelArch;
+        platform_machine = targetMachine;
```
Infinisil, line 98: Does `platform_machine` support being `null`? I think a good alternative is to just throw an error instead of returning `null` for `targetMachine`

K900 (commented):
~~I think I'm stuck in a "started a review" state. Thanks Github. This is my attempt to get myself unstuck.~~

Edit: unstuck success.

pep508.nix
K900: This is probably fine, because this is only used with markers like `platform_machine == "x86_64"`, and `"null"` shouldn't match any of those. The _correct_ way to do it would be to have a complete `uname -m` mapping for every machine type supported by nixpkgs and use that value here...

K900: I think the best solution for now is to just add a comment indicating this and hope no one uses specific dependencies on weird architectures, because mapping every machine type in nixpkgs is definitely a bit too much.

Infinisil (approved):

Infinisil:
CI still broken, but this PR fixes at least the first error of it, so merging

K900:
I was going to add a comment re: platform stuff :( 

Well, guess I'll keep it in mind and try to get it fixed on the nixpkgs side then. Thanks!

Infinisil:
Ah sorry, should be fine, but if you have some comment prepared, feel free to open another PR for it, I can quickly merge it

adisbladis:
Regarding the preexisting CI failures (the `git-deps` test in particular) I've bisected nixpkgs and found https://github.com/NixOS/nixpkgs/commit/b0e372864a4ed1413968b13fa13b7ddf3f4673e8.

I don't really know where to go from here, I don't have the energy to debug this any further.

PR #262, opened 2021-02-20
overrides for gdal, moto, pdal, pyproj, others

rskew:


overrides.nix
```diff
@@ -920,6 +951,22 @@ self: super:
...
   );
 
+  pyproj = super.pyproj.overridePythonAttrs (
+    old: {
+      buildInputs = (old.buildInputs or [ ]) ++
+        [ self.cython ];
+      PROJ_DIR = "${pkgs.proj}";
+      PROJ_LIBDIR = "${pkgs.proj}/lib";
+      # pkgs.proj doesn't export the required headers, so grab them from the source repo
+      PROJ_INCDIR = "${pkgs.fetchFromGitHub {
```
adisbladis, line 961: This needs to be changed to something version independent. We don't want to maintain fixed upstreams in our overrides.

Preferably you'd fix this in nixpkgs to include the relevant headers and only then use it in poetry2nix.
Feel free to CC me on any nixpkgs PR.

adisbladis (changes requested):

overrides.nix
rskew: Turns out `pks.proj.dev` has the headers in it already :tada:

rskew:
Oops, I think I force pushed over your force push!

How are you running `nixpkgs-fmt`? When I run it it creates a huge diff (beyond my changes)

adisbladis:
I think maybe you're on an older version of `nixpkgs-fmt`? It's not been super stable in it's formatting output.

rskew:
That could be it.
Cheers :)

Issue #247, opened 2021-02-01
mkPoetryPackages no longer propagates dependencies


sireliah:
Hello! I'm using `mkPoetryPackages` for building Python dependencies in Poetry project for AWS lambda environment and I noticed that there was a breaking change between `1.12.0` and `1.14.0`. The problem is the `mkPoetryPackages` in `1.12.0` would include the `propagatedBuildInput` of Poetry dependency in `poetryPackages` attribute. In other words `mkPoetryPackages` would flatten the dependency graph (so sub-dependencies would be also included). 

Unfortunately in `1.14.0` this no longer happens.

pyproject.toml
```toml
[tool.poetry]
name = "lambdas"
version = "0.1.0"
description = ""
authors = ["some@author.com"]

[tool.poetry.dependencies]
python = "^3.8"
aws-xray-sdk = "^2.6.0"

[tool.poetry.dev-dependencies]
pytest = "^5.2"
pylint = "2.5.3"
mypy = "^0.790"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

Before:
```bash
Welcome to Nix version 2.3.1. Type :? for help.

nix-repl> pkgs = import <nixpkgs> {}

nix-repl> appPackagesDrv = pkgs.poetry2nix.mkPoetryPackages { projectDir = ./.; }

nix-repl> appPackagesDrv.poetryPackages

[ «derivation /nix/store/9wdcyifqi7ycy6m74sj8w0rrhjpjdcms-python3.8-astroid-2.4.2.drv»
«derivation /nix/store/64q1xxflifqrbnxscvxlr2s31jmw6bh9-python3.8-atomicwrites-1.4.0.drv»
«derivation /nix/store/8abl5cy805q52aajf5rs0cmbzh799za3-python3.8-attrs-20.3.0.drv»
«derivation /nix/store/wygisqa1wf2w043gcn9b20k0r11bl92p-python3.8-aws-xray-sdk-2.6.0.drv»
«derivation /nix/store/bllz9rlkhnnr07yh2jgq3zsrb23yfkhq-python3.8-botocore-1.19.40.drv»
«derivation /nix/store/30h74743ka9dqqzrkdcashsal60h0r7m-python3.8-colorama-0.4.4.drv»
«derivation /nix/store/jyfiiayfzjvj7f3axy7afcask3nlpwfa-python3.8-future-0.18.2.drv»
[...]

PR #270, opened 2021-03-12
Fix mkPoetryPackages: should return deps of deps

sireliah:
This is a fix for [#247 mkPoetryPackages no longer propagates dependencies
](https://github.com/nix-community/poetry2nix/issues/247)

@adisbladis Please take a look and check if this is stable solution!

Important facts:
- I'm doing some string operations on python.pythonVersion to produce right attribute (should look like this: `python37`, `python38`, etc.) and I have no idea if there is better option for that.
- Added tests for current and older version of Python.
- Added documentation how to run tests locally (Tests finish with "Hello, World!" and gunicorn server listening, is this expected?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,8 +329,11 @@ Contributions to this project are welcome in the form of GitHub PRs. Please cons
 - This project uses [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt) for formatting the Nix code. You can use
 `nix-shell --run "nixpkgs-fmt .` to format everything.
 - If you are planning to make any considerable changes, you should first present your plans in a GitHub issue so it can be discussed.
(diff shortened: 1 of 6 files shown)


sireliah:
I'll squash the commits once the review is done. :+1:

adisbladis:
> I'm doing some string operations on python.pythonVersion to produce right attribute (should look like this: python37, python38, etc.) and I have no idea if there is better option for that.

Why is this required? This sounds like something that we shouldn't be concerned with.

sireliah:
You're right, that part could have been just `python.pkgs.requiredPythonModules`.

adisbladis:
Thank you! LGTM now :)

PR #276, opened 2021-03-30
Provide necessary overrides to build `sentence-transformers`

NinjaTrappeur:
More info in the commit msgs.

As you can see, I kinda chickened out for `tokenizers` and used a wheel after seeing the nixpkgs override: https://github.com/NixOS/nixpkgs/blob/f5e8bdd07d1afaabf6b37afc5497b1e498b8046f/pkgs/development/python-modules/tokenizers/default.nix#L50

Is there a way to re-use the nixpkgs source build? If not, I hope the wheel is good enough!

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1560,6 +1560,20 @@ self: super:
     }
   );
 
+  sentencepiece = super.sentencepiece.overridePythonAttrs (
+    old: {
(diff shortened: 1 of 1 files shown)


adisbladis:
> Is there a way to re-use the nixpkgs source build? If not, I hope the wheel is good enough!

No, sadly not.

PR #281, opened 2021-04-07
Add overrides for wxPython

tfmoraes:
It builds wxPython with GTK+3 backend. Tested with the wxPython demo.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1638,50 +1638,52 @@ self: super:
     }
   );
 
-  wxpython = super.wxpython.overridePythonAttrs (old: rec{
-    DOXYGEN = "${pkgs.doxygen}/bin/doxygen";
+  wxpython = super.wxpython.overridePythonAttrs (old:
+    let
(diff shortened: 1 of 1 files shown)


overrides.nix
```diff
@@ -1638,5 +1638,50 @@ self: super:
     }
   );
 
+  wxpython = super.wxpython.overridePythonAttrs (old: rec{
+    DOXYGEN = "${pkgs.doxygen}/bin/doxygen";
+
+    localPython = self.python.withPackages (ps: with ps; [
```
Infinisil, line 1644: Use a `let ... in` instead and remove the `rec`. Without this my initial thought was that `localPython` was some special python package attribute, which is confusing.

Infinisil (changes requested):

overrides.nix
tfmoraes: Ok... Done! Thanks!

Infinisil (approved):
Thanks!

PR #285, opened 2021-04-12
Add overrides for django, django-*, pillow, typed_ast and weasyprint packages.

confususs:
I can split them up if you'd like. Either way, these are some overrides I needed today. 

I'm not sure if I use the correct phase + approach for the `django-stubs-ext` package, let me know if something should be different :smile:.

This was the error I got when I tried to install without this override:
```
  error: [Errno 2] No such file or directory: '../LICENSE.txt'
  ----------------------------------------
  ERROR: Failed building wheel for django-stubs-ext
```

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -178,7 +178,7 @@ self: super:
     super.django.overridePythonAttrs (
       old: {
         propagatedNativeBuildInputs = (old.propagatedNativeBuildInputs or [ ])
-          ++ [ pkgs.gettext ];
+          ++ [ pkgs.gettext self.pytest-runner ];
       }
     )
   );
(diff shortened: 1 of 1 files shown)


overrides.nix
```diff
@@ -193,6 +193,40 @@ self: super:
...
+  django-rosetta = super.django-rosetta.overridePythonAttrs (
+    old: {
+      nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ self.pytest-runner ];
+    }
+  );
+
+  django-stubs-ext = super.django-stubs-ext.overridePythonAttrs (
+    old: {
+      prePatch = ''
+        if ! test -e ../LICENSE.txt; then
```
adisbladis, line 223: Just a minor nit, but this can be simplified to simply `touch ../LICENSE.txt`.

adisbladis (changes requested):

overrides.nix
confususs: Cool! Simplified it :)

PR #289, opened 2021-04-25
Provide flake template for building an app

FRidh:
Provide a basic flake template for building a Python application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,28 +298,12 @@ in pkgs.poetry2nix.mkPoetryApplication {
 
 ## Using the flake
 
-For the experimental flakes functionality we provide poetry2nix as a flake providing an overlay
-to use with [nixpkgs](https://nixos.org/nixpkgs/manual).
+For the experimental flakes functionality we provide _poetry2nix_ as a flake providing an overlay
+to use with [nixpkgs](https://nixos.org/nixpkgs/manual). Additionally, the flake provides
+a flake template to quickly start using _poetry2nix_ in a project:
 
-#### Example
-
-```nix
-{
-    description = "Your flake using poetry2nix";
-
-    inputs.nixpkgs.url = "github:NixOS/nixpkgs";
-    inputs.utils.url = "github:numtide/flake-utils";
-    inputs.poetry2nix-src.url = "github:nix-community/poetry2nix";
-
-    outputs = {nixpkgs, utils, poetry2nix-src, self}: utils.lib.eachDefaultSystem (system: let
-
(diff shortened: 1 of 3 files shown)


templates/app/flake.nix
```diff
@@ -0,0 +1,34 @@
...
+  description = "Application packaged using poetry2nix";
+
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs";
+  inputs.poetry2nix.url = "github:nix-community/poetry2nix";
+
+  outputs = { self, nixpkgs, flake-utils, poetry2nix }:
+    {
+      # Nixpkgs overlay providing the application
+      overlay = nixpkgs.lib.composeManyExtensions [
```
FRidh, line 11: This part I don't like very much, as I think this can lead to the multiple inheritance diamond problem. The alternative is to have the overlay rely on the `poetry2nix` that is part of Nixpkgs, in which case `poetry2nix.overlay` is removed here. That I don't like at all, because the user of this template explicitly points to the `poetry2nix` repo.

However, further down, where the applications are defined, I think it *is

FRidh (commented):
Maybe add the other functions in the template as well and let the user remove what they don't need? What do you think?

templates/app/flake.nix
teto: I think I am missing the idea here but can't we set `overlays` instead ? e.g. why do we need composeManyExtensions

FRidh: Overlays are not propagated. Consider A depends on B, and B on C. B can use C as overlay for itself. If you now want to use B in A through overlays, then you need to use the overlays that B in order to use B.

README.md
```diff
@@ -298,28 +298,12 @@ in pkgs.poetry2nix.mkPoetryApplication {
...
-
-      pkgs = import nixpkgs { inherit system; overlays = [ poetry2nix-src.overlay ]; };
-
-    in
-      {
-         # use pkgs.poetry2nix here.
-      });
-  }
+```sh
+nix flake --template github:nix-community/poetry2nix
```
legendofmiracles, line 306:
```suggestion
nix flake --template github:nix-community/poetry2nix
```
Command doesn't work on my system, and yes i replaced nix-community with your username

teto:
I would be interested to see how to write a proper devShell in the template too, as I fail to do so (see https://github.com/nix-community/poetry2nix/issues/325).

adisbladis:
I would love to see a `devShell` in there, but for now I feel fine with merging it as-is. Still better than not having any template.

PR #297, opened 2021-05-07
safer overrides using 'or [ ]'

NorfairKing:
Every once in a while I get an error because of an override that doesn't use `or [ ]` so I figured I'd go and fix them all with a little find/replace while my code is compiling.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -8,13 +8,13 @@ self: super:
 {
   automat = super.automat.overridePythonAttrs (
     old: rec {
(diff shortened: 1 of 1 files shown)


adisbladis:
Ugh... We should probably have a nice abstraction for this, but in the mean time lets just merge this as it's a clear improvement.

I took the liberty of running `nixpkgs-fmt` and force pushed.

NorfairKing:
@adisbladis thanks!

Issue #309, opened 2021-05-21
Cannot build Python extras with dash in name "-"


sireliah:
Hi everyone! I've found a pretty specific issue with adding a Python extras which contains a dash symbol.

```
poetry add boto3-stubs[cognito-idp] 
```
## The error:
```bash
$ nix build
error: value is null while a list was expected, at /nix/store/sccc9mb8brnc6n2xmnac8hz67sj5qpcm-nixpkgs-21.05pre278688.c0e88185200/nixpkgs/pkgs/development/tools/poetry2nix/poetry2nix/pep508.nix:135:19
(use '--show-trace' to show detailed location information)
```
## The cause
[The regex](https://github.com/nix-community/poetry2nix/blob/master/pep508.nix#L132) that is parsing the submodules seems to be missing the dash.

## The possible fix
Adding the dash to the regex fixes the issue.
```
mVal = ''[a-zA-Z0-9\'"_\. \-]+'';
```

## Test setup:

nix (Nix) 2.3.1
channel: https://nixos.org/channels/nixpkgs-unstable
Poetry version: 1.1.6
poetry2nix version: master

```toml
# pyproject.toml
[tool.poetry]
name = "nix-poetry-test"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "^3.8"
boto3-stubs = {extras = ["cognito-idp"], version = "^1.17.77"}

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

```nix
# default.nix
{ pkgs ? import <nixpkgs> {} }:

let
  poetry2nix = pkgs.poetry2nix;
  python3 = pkgs.python3;
in
poetry2nix.mkPoetryApplication {
  python = python3;
  projectDir = ./.;
}
```

If you're fine with that fix, I'll open a PR. Thanks for the attention!

sireliah:
Solved!

PR #312, opened 2021-05-27
Allow Python extras to contain dashes

sireliah:
The regex for building Python extras (optional sub-dependencies) was not allowing dashes in names, such as:

`poetry add 'boto3-stubs[cognito-idp]'`

For detailed explanation what was missing, please see: https://github.com/nix-community/poetry2nix/issues/309

How to test:

```bash
cd ./tests/extras

# Build test locally
nix-shell -E 'with import <nixpkgs> {}; let p = callPackage ../../default.nix {}; in callPackage ./default.nix {poetry2nix = p;}'
```
This command should not fail.


tests/extras/pyproject.toml
```diff
@@ -8,6 +8,9 @@ authors = ["Your Name <you@example.com>"]
 python = "^3.6"
 setuptools-scm = { version = "^3.4", extras = ["toml"] }
 
+# Extra with dash in the name should work
+boto3-stubs = {extras = ["cognito-idp"], version = "^1.17.82"}
```
sireliah, line 12: If you have better idea for package that could be used for testing here, I'm open for suggestions!

PR #321, opened 2021-06-08
Add overrides for marisa-trie, ua-parser and lazy-object-proxy

daedric:


overrides.nix
```diff
@@ -1839,4 +1839,16 @@ self: super:
...
 
+  marisa-trie = super.marisa-trie.overridePythonAttrs (
+    old: {
+      buildInputs = (old.buildInputs or [ ]) ++ [ self.pytest-runner ];
+    }
+  );
+
+  ua-parser = super.ua-parser.overridePythonAttrs (
+    old: {
+      buildInputs = (old.buildInputs or [ ]) ++ [ self.pyyaml ];
```
adisbladis, line 1850: Are you sure this one shouldn't be propagated?

daedric: Yes you're right. I did not check where pyyaml was used and it is used outside of the setup.py as well.
I'm updating the PR

Issue #324, opened 2021-06-16
Can't install pendulum 2.1.2 on Linux


zupo:
Installing latest pendulum 2.1.2 on Linux fails with `ERROR: pendulum-2.1.2-cp38-cp38-manylinux_2_32_x86_64.whl is not a supported wheel on this platform.`.

The workaround I am currently using is this:

```
  poetryOverrides = self: super: {
    pendulum = super.pendulum.override {
      preferWheel = true;
    };
  };
```

I've prepared a Minimal Reproducible Example that showcases the problem: https://github.com/niteoweb/big-sur-poetry-test/tree/pendulum

Surprisingly, building works fine on MacOS, maybe due to lack of sandboxing?

NorfairKing:
I'm trying to upstream this workaround including a regression test, but I can't get the regression test to fail ..?

adisbladis:
This is now fixed.

zupo:
Confirmed it works in my example too, thanks!

https://github.com/niteoweb/big-sur-poetry-test/commit/e64e2e52476c0cdccee25f28d2cee52debb8b5d3

PR #326, opened 2021-06-26
Fix for pendulum

NorfairKing:
This is a fix for #324 

The first commit is a failing regression test, the rest is a fix.


NorfairKing:
Definitely squash this before merging, so the circleci file doesn't end up in the git history.

adisbladis:
It seems like the root cause of this issue is https://www.python.org/dev/peps/pep-0600/ + some bug in pip I've yet to identify where the new pep-600 wheel identifiers are not accepted for whatever reason.

Maybe @FRidh knows what's happening here?

PR #329, opened 2021-06-30
Fix insecure ssl in fetch_from_legacy

nlewo:
When the first PR has been submitted, I think the store path was already in my cache and in the CI cache so tests haven't been able to detect this error (moreover, I tried to clean my `/nix/store` but it seems i failed:/).


adisbladis:
Thanks!

PR #330, opened 2021-07-01
Add overrides for multiple packages

jgeerds:
This pull request fixes multiple packages:

- pygraphviz
- pyjsg
- pyshex
- pyshexc
- shexjsg
- sparqlslurper

Unfortunately, I think there are lot of packages depend on `pbr`. Is there any way to automatically detect the `pbr` requirement and apply the fix to the derivation?

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1865,4 +1865,28 @@ self: super:
     '';
   });
 
+  pygraphviz = super.pygraphviz.overridePythonAttrs (old: {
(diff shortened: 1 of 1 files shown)


adisbladis:
For now we have to rely on adding manual overrides because of https://github.com/python-poetry/poetry/issues/2789.
Ideally this would be solved upstream and we'd be able to pick up on that.

takeda:
@adisbladis yay, it just got merged

PR #335, opened 2021-07-20
Fix matplotlib build

utdemir:
Currently, building matplotlib fails with error:

```
ERROR: Could not find a version that satisfies the requirement certifi>=2020.06.20 (from versions: none)
```

This PR adds the failing test case, and fixes it by inserting the `certifi` dependency.

Let me know if there is a better solution.


adisbladis:
LGTM. Thanks!

PR #342, opened 2021-08-03
Enable dataset module in pyarrow

lunik1:
I think this is worth enabling as it is used by dask when loading data from parquet files. You get `ModuleNotFoundError: No module named 'pyarrow._dataset'` at runtime if you try without having this enabled.


adisbladis:
Thank you!

PR #349, opened 2021-08-11
Bump version to 1.18.0

nlewo:
Because it would be nice to have a new release in nixpkgs (i mainly interested by a `fetch-from-legacy` fix).

(Once released, I could then submit a PR in nixpkgs.)


adisbladis:
Tagged!

PR #356, opened 2021-08-17
Add override for tomli

trepetti:


overrides.nix
```diff
@@ -1907,4 +1907,8 @@ self: super:
   sparqlslurper = super.sparqlslurper.overridePythonAttrs (old: {
     buildInputs = (old.buildInputs or [ ]) ++ [ self.pbr ];
   });
+
+  tomli = super.tomli.overridePythonAttrs (old: {
+    buildInputs = (old.buildInputs or [ ]) ++ [ self.flit-core ];
```
adisbladis, line 1912: Since this is a `build-system` input it should be `nativeBuildInputs`.

adisbladis (changes requested):

overrides.nix
trepetti: So this is something that confuses me a bit. `tomli` as packaged already has `flit-core` in its `nativeBuildInputs` by default in `nixpkgs`, but will fail with an import error as is. I noticed that `packaging`, which suffered from a similar issue, also always had `flit-core` in its `nativeBuildInputs` ([see here](https://github.com/NixOS/nixpkgs/commit/8149718529d6faa7398aa4553c1301029920daed#diff-b0a56515ffa90867b056e9602299857b8149a3c7b21340d13ff0275e798e4881)) but added an override to add `flit-core` to `buildInputs` to fix this behavior when being used within `poetry2nix`. I definitely agree in principle that it should be a `nativeBuildInputs` component, but in practice it looks like the module will only be available if it is in `buildInputs`.

trepetti: This is the relevant PR that added this same build system dep for `packaging`, also putting it in `buildInputs`.

adisbladis: Fair enough. Not going to debug this issue any further right now so I'm fine with this as-is.

takeda: @adisbladis I run into this accidentally, because I'm having this issue, but looks like my version is just older than this fix.

I'm wondering if nativeBuildInputs actually has any meaning in case of Python. The language is interpreted so technically whatever runs through python is already running on the target system.

@trepetti mentioned that filt-core is already in `nativeBuildInputs`. I did not verify that, but perhaps poetry2nix should be populating those into `buildInputs` so overrides like this wouldn't be necessary?

The thing is that it is still a bit hairy, for example `cython` makes more sense to be in `nativeBuildInputs` since it generates code. I think though more times than not `buildInputs` would make sense :/. What do you think?

trepetti: There are many other overrides in here for `flit-core` and other alternate build systems that packages rely on. The fix did not jump out at me when I looked through `poetry2nix` code, but yes, that is the gist of the problem. You may have better luck than I did, though.

PR #366, opened 2021-08-25
feat: Make pytest-randomly installable

### LINKED ISSUE #322: No matching distribution found for importlib-metadata>=3.6.0 (from versions: none)
I get the following error when installing `pytest-randomly` and `ruamel-yaml`:

```
# ERROR: Could not find a version that satisfies the requirement ruamel.yaml.clib>=0.1.2; platform_python_implementation == "CPython" and python_version < "3.10" (from ruamel.yaml==0.16.13) (from versions: none)
# ERROR: No matching distribution found for ruamel.yaml.clib>=0.1.2; platform_python_implementation == "CPython" and python_version < "3.10" (from ruamel.yaml==0.16.13)
```

I use this workaround to make it work on my local:

```
  poetryOverrides = self: super: {

    ruamel-yaml = super.ruamel-yaml.overrideAttrs (old: {
      postPatch = old.postPatch or "" + ''
        substituteInPlace __init__.py \
          --replace "ruamel.yaml.clib>=0.1.2" ""
      '';
    });

    pytest-randomly = super.pytest-randomly.overrideAttrs (old: {
      postPatch = old.postPatch or "" + ''
        sed -i 's/importlib-metadata >= 3.6.0 ; python_version < "3.10"//' setup.cfg
      '';
    });

  };
```

Ideally I'd like to make a proper fix for this and upstream it, but I'm not sure what exactly is going on and how to fix it.

l0b0:
Based on
<https://github.com/nix-community/poetry2nix/issues/322#issue-922199285>.

Closes #322 together with #365.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1285,6 +1285,12 @@ self: super:
     }
   );
 
+  pytest-randomly = super.pytest-randomly.overrideAttrs (old: {
+    postPatch = old.postPatch or "" + ''
+      sed -i 's/importlib-metadata >= 3.6.0 ; python_version < "3.10"//' setup.cfg
+    '';
+  });
+
   pytest-runner = super.pytest-runner or super.pytestrunner;
 
   pytest-pylint = super.pytest-pylint.overridePythonAttrs (


cpcloud:
This patch seems to break use of `pytest-randomly` on Pythons != 3.7.

cpcloud:
Here's a failing test in a PR to reproduce the issue https://github.com/nix-community/poetry2nix/pull/387/

l0b0:
@cpcloud Thank you for the PR; the issue was noted in https://github.com/nix-community/poetry2nix/issues/376 and the follow-up https://github.com/nix-community/poetry2nix/issues/379, but I'm not sure what the correct fix should be.

cpcloud:
I think we should probably revert the fix that removes the importlib metadata dependency, since the library still depends on it.

l0b0:
I tried to revert by changing the override locally to

```nix
pytest-randomly = super.pytest-randomly.overrideAttrs (
  old: {
    postPatch = "";
    propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ self.importlib-metadata ];
  }
);
```

, but that didn't work:

```console
$ pytest
[…]
ModuleNotFoundError: No module named 'importlib_metadata'
```

cpcloud:
`pytest-randomly` needs the override, not `pytest`

l0b0:
That's what I thought, but this is what works for me when installing both packages.

cpcloud:
Did you also remove the `postPatch`? Here's a project where I'm overriding only `pytest-randomly`: https://github.com/cpcloud/stupidb/blob/master/poetry-overrides.nix

l0b0:
If I do just

```nix
        pytest-randomly = super.pytest-randomly.overrideAttrs (attrs: {
          propagatedBuildInputs = (attrs.propagatedBuildInputs or [ ]) ++ [ self.importlib-metadata ];
          postPatch = "";
        });
```

then `pytest` still reports "ModuleNotFoundError: No module named 'importlib_metadata'". If I instead do

```nix
        pytest = super.pytest.overrideAttrs (
          old: {
            propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ self.importlib-metadata ];
          }
        );
        pytest-randomly = super.pytest-randomly.overrideAttrs (attrs: {
          postPatch = "";
        });
```

then `pytest` works.

PR #387, opened 2021-09-16
python: fix marker comparison

### LINKED ISSUE #379: `markers = "python_version < \"3.10\""` treated like 3.1?
It looks like `markers = "python_version < \"3.10\""` is compared to the current Python version numerically, because removing it, for example from `importlib-metadata = {version = ">=3.6.0", markers = "python_version < \"3.10\""}`, turns a

> Could not find a version that satisfies the requirement importlib-metadata>=3.6.0; python_version < "3.10" (from pytest-randomly==3.10.1) (from versions: none)

error into no error. Another workaround is to change the marker into `"python_version < \"3.9.9999\"`.

Removing the marker from setup.cfg does _not_ help:

```nix
pytest-randomly = super.pytest-randomly.overrideAttrs (old: {
  postPatch = ''
    sed -i 's/importlib-metadata >= 3.6.0 ; python_version < "3.10"/importlib-metadata >= 3.6.0/' setup.cfg
  '';
});
```

Presumably because the poetry.lock restriction is also taken into account when trying to resolve a version.

This may be the underlying cause of https://github.com/nix-community/poetry2nix/issues/376.

In the meantime, is it possible to override the contents of poetry.lock? I don't see any other overrides doing that.

cpcloud:
This PR is a possible fix for #379, using version comparison functions instead of comparison operators for comparing marker values.

A previously failing test is included.

Closes #379.

### DIFF
--- a/pep508.nix
+++ b/pep508.nix
@@ -160,12 +160,12 @@ let
       hasElem = needle: haystack: builtins.elem needle (builtins.filter (x: builtins.typeOf x == "string") (builtins.split " " haystack));
       op = {
         "true" = x: y: true;
-        "<=" = x: y: (unmarshal x) <= (unmarshal y);
-        "<" = x: y: (unmarshal x) < (unmarshal y);
+        "<=" = x: y: lib.versionAtLeast (unmarshal y) (unmarshal x);
+        "<" = x: y: lib.versionOlder (unmarshal x) (unmarshal y);
         "!=" = x: y: x != y;
         "==" = x: y: x == y;
-        ">=" = x: y: (unmarshal x) >= (unmarshal y);
-        ">" = x: y: (unmarshal x) > (unmarshal y);
+        ">=" = x: y: lib.versionAtLeast (unmarshal x) (unmarshal y);
+        ">" = x: y: lib.versionOlder (unmarshal y) (unmarshal x);
         "~=" = v: c:
           let
             parts = builtins.splitVersion c;
--- a/pep508.nix
+++ b/pep508.nix
@@ -160,12 +160,12 @@ let
       hasElem = needle: haystack: builtins.elem needle (builtins.filter (x: builtins.typeOf x == "string") (builtins.split " " haystack));
       op = {
         "true" = x: y: true;
-        "<=" = x: y: lib.versionAtLeast (unmarshal y) (unmarshal x);
+        "<=" = x: y: op.">=" y x;
         "<" = x: y: lib.versionOlder (unmarshal x) (unmarshal y);
         "!=" = x: y: x != y;
         "==" = x: y: x == y;
         ">=" = x: y: lib.versionAtLeast (unmarshal x) (unmarshal y);
-        ">" = x: y: lib.versionOlder (unmarshal y) (unmarshal x);
+        ">" = x: y: op."<" y x;
         "~=" = v: c:
           let
             parts = builtins.splitVersion c;
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -28,6 +28,7 @@ builtins.removeAttrs
   top-packages-2 = callTest ./common-pkgs-2 { };
   pep425 = pkgs.callPackage ./pep425 { inherit pep425;inherit pep425OSX;inherit pep425Python37; };
   env = callTest ./env { };
+  pytest-randomly = callTest ./pytest-randomly { };
   file-src-deps = callTest ./file-src-deps { };
   file-src-deps-level2 = callTest ./file-src-deps-level2 { };
   file-wheel-deps = callTest ./file-wheel-deps { };
--- a/tests/pytest-randomly/default.nix
+++ b/tests/pytest-randomly/default.nix
@@ -1,0 +1,13 @@
+{ lib, poetry2nix, python39, runCommand }:
+let
+  env = poetry2nix.mkPoetryEnv {
+    python = python39;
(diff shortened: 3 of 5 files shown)


cpcloud:
Looks like this could be caused by https://github.com/nix-community/poetry2nix/issues/379.

cpcloud:
@l0b0 I've added a possible fix for #379 here. Can you review?

pep508.nix
```diff
@@ -160,12 +160,12 @@ let
       hasElem = needle: haystack: builtins.elem needle (builtins.filter (x: builtins.typeOf x == "string") (builtins.split " " haystack));
       op = {
         "true" = x: y: true;
-        "<=" = x: y: (unmarshal x) <= (unmarshal y);
-        "<" = x: y: (unmarshal x) < (unmarshal y);
+        "<=" = x: y: op.">=" y x;
```
l0b0, line 163: Would `op."<=" x y;` be clearer?

pep508.nix
```diff
@@ -160,12 +160,12 @@ let
...
+        "<" = x: y: lib.versionOlder (unmarshal x) (unmarshal y);
         "!=" = x: y: x != y;
         "==" = x: y: x == y;
-        ">=" = x: y: (unmarshal x) >= (unmarshal y);
-        ">" = x: y: (unmarshal x) > (unmarshal y);
+        ">=" = x: y: lib.versionAtLeast (unmarshal x) (unmarshal y);
+        ">" = x: y: op."<" y x;
```
l0b0, line 168: Same here - would `op.">" x y` be clearer?

pep508.nix
cpcloud: I think that would cause infinite recursion right?

l0b0: Oh, is `op.">="` the same as `>=`? In that case I'm not sure what the final bindings end up as. We might want a bunch more tests to make sure they all work as expected.

l0b0:
You seem quite familiar with Nix - do you know how to create a poetry2nix which applies your patches? I tried the following (I've not yet fetched the checksums), but `nix-shell` complains about "attribute 'overrideAttrs' missing":

```nix
  poetry2nix = pkgs.poetry2nix.overrideAttrs (old: {
    patches = (old.patches or [ ]) ++ [
      (
        pkgs.fetchpatch
          {
            url = "https://github.com/cpcloud/poetry2nix/commit/38c2f51e8583e0104ea5d502f2e4c583fb5d5108.patch";
            sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
          }
          pkgs.fetchpatch
          {
            url = "https://github.com/cpcloud/poetry2nix/commit/b6f569db2af5998bed5626966e152473c5360372.patch";
            sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
          }
          pkgs.fetchpatch
          {
            url = "https://github.com/cpcloud/poetry2nix/commit/5b7ffec522f92ac3992feb723fb8408c8d6d5c98.patch";
            sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
          }
      )
    ];
  });
```

cpcloud:
cc @adisbladis for additional review

pep508.nix
cpcloud: I think it's not as complex as you might think. What's happening here is analogous to the following code in `nix repl`:

```
nix-repl> let op = { gt = x: y: x > y; lt = x: y: op.gt y x; }; in op.lt 1 2
true
```

cpcloud:
> You seem quite familiar with Nix - do you know how to create a poetry2nix which applies your patches? I tried the following (I've not yet fetched the checksums), but `nix-shell` complains about "attribute 'overrideAttrs' missing":
> 
> ```nix
>   poetry2nix = pkgs.poetry2nix.overrideAttrs (old: {
>     patches = (old.patches or [ ]) ++ [
>       (
>         pkgs.fetchpatch
>           {
>             url = "https://github.com/cpcloud/poetry2nix/commit/38c2f51e8583e0104ea5d502f2e4c583fb5d5108.patch";
>             sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
>           }
>           pkgs.fetchpatch
>           {
>             url = "https://github.com/cpcloud/poetry2nix/commit/b6f569db2af5998bed5626966e152473c5360372.patch";
>             sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
>           }
>           pkgs.fetchpatch
>           {
>             url = "https://github.com/cpcloud/poetry2nix/commit/5b7ffec522f92ac3992feb723fb8408c8d6d5c98.patch";
>             sha256 = "07sx98d422589gxr8wflfpkdd0k44kbagxl3b51i56ky2wfix7rc";
>           }
>       )
>     ];
>   });
> ```

@l0b0 I'm not entirely sure why you're getting that error (I get it too). It looks like maybe `poetry2nix` isn't producing an overridable derivation. I personally use `niv` to manage my nix projects' dependencies, and pull in `poetry2nix` from this repository, which allows me to test things out from any GitHub fork (including my own).

l0b0:
Now that the comparison is fixed, can we possibly get rid of a bunch of overrides? Basically anything which failed to install because of string comparison of the version is now obsolete.

Issue #375, opened 2021-08-30
Can't parse "platformdirs>=2" in setup.py


l0b0:
When trying to install the latest version of Black I get the following message:

```
ERROR: Could not find a version that satisfies the requirement platformdirs>=2 (from black==21.8b0) (from versions: none)
ERROR: No matching distribution found for platformdirs>=2 (from black==21.8b0)
```

Workaround:

```
black = super.black.overridePythonAttrs (old: {
  postPatch = ''
    substituteInPlace setup.py --replace 'platformdirs>=2' 'platformdirs'
  '';
});
```

PR #388, opened 2021-09-17
feat: Make Black installable

l0b0:
Closes https://github.com/nix-community/poetry2nix/issues/375.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -94,6 +94,13 @@ self: super:
     }
   );
 
+  # https://github.com/nix-community/poetry2nix/issues/375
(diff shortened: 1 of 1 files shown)


cpcloud:
Should we try to figure out why the parsing fails instead of removing the constraint?

cpcloud:
This should probably also have a test to avoid a regression, if possible.

l0b0:
You're absolutely right @cpcloud. I've just been pushing out PRs to be able to move on from being completely blocked, but I never looked into the underlying issue.

adisbladis:
Closing in favour of #398.

PR #390, opened 2021-09-17
Enable private pypi repositories with credentials

rskew:
I've been wanting this for a while, so I finally had a crack at it.

Credentials are declared in a file which is passed to poetry2nix as an argument.

I used this project to test it out: https://github.com/rskew/poetry2nix-private-pypi-demo

I haven't given a lot of thought to the design, I just wanted to draft something that works as a demo. Feedback appreciated!

### DIFF
--- a/fetch_from_legacy.py
+++ b/fetch_from_legacy.py
@@ -4,7 +4,9 @@
 # Note it is not possible to use pip
 # https://discuss.python.org/t/pip-download-just-the-source-packages-no-building-no-metadata-etc/4651/12
 
+import os
 import sys
+import netrc
 from urllib.parse import urlparse, urlunparse
 from html.parser import HTMLParser
 import urllib.request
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -122,6 +122,7 @@ lib.makeScope pkgs.newScope (self: {
       # Example: { my-app = ./src; }
     , editablePackageSources ? { }
     , __isBootstrap ? false  # Hack: Always add Poetry as a build input unless bootstrapping
+    , repositoryCredentialsFile ? null
```
adisbladis, line 125: I don't really like the explicit passing of credentials. I would rather we use `impureEnvVars` https://nixos.org/manual/nix/stable/#sec-advanced-attributes or make the fetcher netrc aware.

lovesegfault:
@rskew Figured I'd check in on this, are you planning on addressing the pending comment / pursuing this further?

Thanks!

default.nix
rskew: Thanks, I wasn't aware of those options.
I've updated it to use netrc, but it has to be passed in via both `extra-sandbox-paths` and an environment var to point to it. I struggled with `impureEnvVars` due to building via the daemon, so I grabbed the variable from the nix path as done in fetchdocker:
- https://github.com/NixOS/nixpkgs/blob/a0dbe47318bbab7559ffbfa7c4872a517833409f/pkgs/build-support/fetchdocker/credentials.nix

I can now install packages from a private pypi via `nix-shell -I NETRC=/etc/nix/.netrc --option extra-sandbox-paths /etc/nix
`, without explicitly passing credentials through the derivation.

Have I missed anything else basic?

rskew:
Thanks for the reminder @lovesegfault , apologies for the delay!

PR #394, opened 2021-09-18
fix: only change shellingham format to pyproject if format is not wheel

jashandeep-sohi:
When building a Poetry application with `preferWheels = true` and  `shellingham` as a dependency, the override changes the format to `pyproject` but the input to derivation is a `.whl` file. Because the format now is `pyproject`, the wheel hook doesn't run and the build fails.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1606,17 +1606,6 @@ self: super:
 
   });
 
-  shellingham =
-    if lib.versionAtLeast super.shellingham.version "1.3.2" then
-      (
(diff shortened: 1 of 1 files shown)


adisbladis:
I believe this override could be dropped altogether as the default format is now `pyproject`?

Could you try simply dropping the override altogether?

jashandeep-sohi:
Ah okay, was not aware that's the default. Yep, can confirm removing the override also works.

adisbladis:
Great! Thank you!

PR #395, opened 2021-09-19
fix: prevent infinite recursion when using tomli

cpcloud:
This prevents infinite recursion against nixos-unstable-small as of NixOS/nixpkgs@ab126c9a62ca414a933ac19e91d27ddd68cedcd4.


cpcloud:
I'm not entirely sure how to test this, since it requires a very recent `nixpkgs/nixos-unstable-small`.

PR #396, opened 2021-09-19
fix: Make pytest runnable with pytest-randomly

l0b0:
Cc @cpcloud, @zupo, @adisbladis.

Refer to discussion on https://github.com/nix-community/poetry2nix/pull/366.


l0b0:
Should no longer be needed after https://github.com/nix-community/poetry2nix/pull/387.

PR #403, opened 2021-09-27
README: recommend overriding env over mkShell

### LINKED ISSUE #186: Add doc on how to create a shell that includes dev-dependencies

ryneeverett:
Resolve #186.

The current documentation shows how to include dev-dependencies in
environment *or* add external dependencies to the environment, but I
suspect in the vast majority of cases both will be desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ in myAppEnv.env
 ```
 
 #### Example shell.nix with external dependencies
-For a shell environment including external dependencies, pass the app environment and dependency packages (for example, `pkgs.hello`) as build inputs to `pkgs.mkShell`.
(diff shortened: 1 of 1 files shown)


adisbladis:
I don't understand this PR. Both of these examples looks functionally equivalent to me?

PR #412, opened 2021-09-30
fix(zipp): skip patching setup.py if format is wheel

jashandeep-sohi:
I'm running into another build issue with `preferWheels = true` and `zipp==3.6.0`. Skipping the `setup.py` patch if format is wheel seems to work.


adisbladis:
I took the liberty of fixing the evaluation error and squashing your commits. Fix LGTM.

Issue #74, opened 2020-03-26
Install the current project


jtrakk:
I'm working on a package that uses Poetry. The source code is in `src/` so that it isn't importable without installation. This is a common project layout (for example, attrs and flask have it). With `mkPoetryEnv` I get access to the dependencies but the project I'm actually working on is not installed so I can't import it. 

The normal solution is to do an editable install, but pip/setuptools [doesn't have support for this yet](https://github.com/pypa/packaging-problems/issues/256#issuecomment-559250309) with Poetry projects. Poetry itself does its own internal implementation to make it work. 

An alternative solution would be to actually install the whole package. This has the advantage that there's no possibility of incomplete installs or other weirdness, and doesn't need any special support from setuptools. The downside is that installing a package after every edit would be slow.

A hacky solution that I can use currently is just setting `PYTHONPATH=src` while I'm working on my project.

adisbladis:
Is there a good way for us to figure out what source directory to add to `PYTHONPATH` automatically?
If there is we could potentially do that in a shell-hook.

jtrakk:
The directory is either `projectDir` or `projectDir + "/src"`, preferring the latter if it exists.

It'll be more robust if poetry2nix can avoid using `PYTHONPATH` and instead use `buildPythonEnv` directly. That might require some thought to figure out exactly the right way to do it. In the meantime, setting `PYTHONPATH` will work fine in most cases though.

mjlbach:
I've also started adding an export to .envrc to put the current project in the PYTHONPATH. I believe this same situation applies with any local module installed via poetry2nix, which in poetry defaults to develop/editable mode.

adisbladis:
@jtrakk or @mjlbach  Do you have an example project for me that has this layout?

jtrakk:
- https://github.com/python-attrs/attrs
- https://github.com/pallets/flask
- https://github.com/pyca/cryptography

Blog post about why src layout is good: https://hynek.me/articles/testing-packaging/

adisbladis:
@jtrakk I should have clarified: With this structure _and_ using Poetry.

PR #416, opened 2021-10-01
Add projectDir to import path for scripts packages

adisbladis:
This ensures that packages retain their editable property despite not being in the current working directory.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -105,7 +105,7 @@ lib.makeScope pkgs.newScope (self: {
     }:
       assert scripts != { };
       import ./shell-scripts.nix {
-        inherit lib python scripts;
+        inherit lib python scripts projectDir;
       };
 
   /*
@@ -130,7 +130,7 @@ lib.makeScope pkgs.newScope (self: {
       scripts = pyProject.tool.poetry.scripts or { };
       hasScripts = scripts != { };
       scriptsPackage = self.mkPoetryScriptsPackage {
-        inherit python scripts;
+        inherit python scripts projectDir;
       };
 
       hasEditable = editablePackageSources != { };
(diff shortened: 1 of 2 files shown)


adisbladis:
I'm not entirely convinced that this is correct to do, since it's not strictly reproducible (the local path will vary).
But the UX implications trump purity in this instance?

I would love some input on this.

cpcloud:
I've been using `PYTHONPATH = builtins.toPath ./.;` in `devShell` (or `mkShell` in `shell.nix`) for this and it works well. Not sure we should force opt-in here.

adisbladis:
> I've been using PYTHONPATH = builtins.toPath ./.; in devShell (or mkShell in shell.nix) for this and it works well. Not sure we should force opt-in here.

Using `$PYTHONPATH` is very impure and is going to impact _any_ Python subprocess without any regards to compatibility, so it's not a solution I would recommend.

cpcloud:
> > I've been using PYTHONPATH = builtins.toPath ./.; in devShell (or mkShell in shell.nix) for this and it works well. Not sure we should force opt-in here.
> 
> Using `$PYTHONPATH` is very impure and is going to impact _any_ Python subprocess without any regards to compatibility, so it's not a solution I would recommend.

Right, I'm aware of the problems with `PYTHONPATH`. It's a viable workaround, and at the same time not recommended.

@adisbladis Can you describe the use case for adding `projectDir` to the import path for scripts packages? Knowing what problem the solution would solve might help clarify what other approaches are reasonable/possible/viable.

adisbladis:
> Can you describe the use case for adding projectDir to the import path for scripts packages? Knowing what problem the solution would solve might help clarify what other approaches are reasonable/possible/viable.

First, assume I'm using `direnv` + a `shell.nix` calling out to `mkPoetryEnv`.
This projects `pyproject.toml` has a set of scripts defined, in this case `some_script`.

The workflow can roughly be described as:
- `$ cd some_pyproject`  <- Activates the environment
- `$ cd some_subdir` <- Environment still active
- `$ some_script`  <- As we are not relative to the root of the this results in an `ImportError.`

adisbladis:
The more I think about this the more sure I am that this  isn't the correct solution.
I think the solution to both this and #74 would be to automatically do an editable install into the environment.

PR #424, opened 2021-10-20
fetchGit: work-around bug in Nix 2.4

vlaci:
Another approach to make `poetry2nix` work with dependencies specified
by arbitrary `rev=<hash>` values on newer Nix revisions.

Relates-to: https://github.com/nix-community/poetry2nix/pull/358
Relates-to: https://github.com/NixOS/nix/issues/5128

--
~Interestingly I only see this error on 'nix-2.4pre20211001_4f49615' (nixpkgs-unstable channel) but not on 'nix-2.5pre20211007_844dd90' (nixos-unstable channel)~
**Update**: there was a cached copy of that dependency in my /nix/store :upside_down_face:

### DIFF
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -163,11 +163,18 @@ pythonPackages.callPackage
       src =
         if isGit then
           (
-            builtins.fetchGit {
+            builtins.fetchGit ({
               inherit (source) url;
               rev = source.resolved_reference or source.reference;
               ref = sourceSpec.branch or sourceSpec.rev or (if sourceSpec?tag then "refs/tags/${sourceSpec.tag}" else "HEAD");
-            }
+            } // (
+              let
+                nixVersion = builtins.substring 0 3 builtins.nixVersion;
+              in
+              lib.optionalAttrs (lib.versionAtLeast nixVersion "2.4") {
+                allRefs = true;
+              }
+            ))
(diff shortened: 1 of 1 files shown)


vlaci:
I was mistaken, the issue persists on 2.5 as well, so the condition check need to be adjusted accordingly

adisbladis:
This bug should be fixed in Nix, not worked around in `poetry2nix`.

vlaci:
I'd like you to reconsider this decision because unless I am mistaken there is no alternative right now to use commit hashes with flakes.
There is a similar workaround for yarn2nix in nixpkgs proper https://github.com/NixOS/nixpkgs/pull/141005/files

mou:
I agree with @vlaci. But for me it's not a workaround, it's plain following to upstream API changes

adisbladis:
I couldn't force-push to this branch so I made a new PR in #512.

adisbladis:
> it's plain following to upstream API changes

The point is that we _shouldn't have_ these types of regressions in Nix and force downstream projects to deal with them.

mou:
@adisbladis i agree, Nix team should communicate clearly and announce breaking changes in advance. Downstream projects should have time to react and adapt. In a week of learning Nix this is already second case when upstream make dependent projects broken. It's not ok ((

PR #433, opened 2021-11-08
fix(pip): remove version restriction

### LINKED ISSUE #244: CI fails with nixpkgs after pip 20.3 update
After bisecting this, I found that CI started failing with the pip 20.2.4 -> 20.3 update in https://github.com/NixOS/nixpkgs/commit/b0e372864a4ed1413968b13fa13b7ddf3f4673e8. The error is in regards to git dependencies:

```
Collecting alembic@ git+https://github.com/sqlalchemy/alembic.git@rel_1_3_1
  Cloning https://github.com/sqlalchemy/alembic.git (to revision rel_1_3_1) to /build/pip-install-i48hvlvt/alembic_3450e6726d6b44b6ab41a7e579ae0c67
  ERROR: Error [Errno 2] No such file or directory: 'git' while executing command git clone -q https://github.com/sqlalchemy/alembic.git /build/pip-install-i48hvlvt/alembic_3450e6726d6b44b6ab41a7e579ae0c67
ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?
builder for '/nix/store/df9ciykjrpyym58kc88vlcdqdjaq6sii-python3.8-git-deps-0.1.0.drv' failed with exit code 1
```

For some reason pip doesn't detect the passed dependency anymore and tries to fetch it on its own. This error is reproducible with
```
$ nix-build tests -A git-deps -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/master.tar.gz
```

cpcloud:
This PR addresses #244.

Problem(s):
1. pip >=20.3 unconditionally clones git dependencies
1. pip >=21 unconditionally clones git dependencies and provides no way to disable the behavior

Solution:

Adopt the same approach as path dependencies, and remove the `git` (and any associated keys) key and
insert a `version = "*"` specification.

Since git dependencies have already been fetched and built with the specific hash by `poetry2nix`, any
version can be used, since (I think) the only version that can get picked up is the desired one.

Additional changes:

1. Unification of git and path dependency handling into a single Python script.

Additional notes:

1. Unfortunately pip >=20.3 hardcodes a list of a few packages it considers "never installed", because
   they overlap with stdlib modules. This prevents the `argparse` dependency from ever resolving.
   The solution is to constrain the dependency on the Python version (see https://github.com/pypa/pip/issues/9035).
   I think this also means that this is a major breaking change for `poetry2nix`.
1. I've also pinned `cryptography = "^3"` in the `common-pkgs-2` test, otherwise we have to override `cryptography`
   and use `cargoSetupHook`, which itself requires bumping `nixpkgs`
   (`cargoSetupHook` is newish). I thought it would be easier to review if
   there were fewer unrelated changes in the PR. I can bump cryptography in a follow up.

Closes #244.


cpcloud:
cc @adisbladis If you could review whenever you get a chance :) This PR should also largely unblock use of Python 3.10.

adisbladis:
Looking very good!

Thanks :tada:

PR #440, opened 2021-11-18
fix(overrides): add flit-core to typing-extensions

lovesegfault:
Since 4.0.0 they use flit for the build

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1763,6 +1763,13 @@ self: super:
   # nix uses a dash, poetry uses an underscore
   typing_extensions = super.typing_extensions or self.typing-extensions;
 
+  typing-extensions = super.packaging.overridePythonAttrs (
+    old: {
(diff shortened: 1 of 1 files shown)


rissson (approved):

overrides.nix
```diff
@@ -1763,6 +1763,13 @@ self: super:
   # nix uses a dash, poetry uses an underscore
   typing_extensions = super.typing_extensions or self.typing-extensions;
 
+  typing-extensions = super.packaging.overridePythonAttrs (
```
takeda, line 1766: Shouldn't it be `super.typing-extensions.overridePythonAttrs` ?

cpcloud: @lovesegfault Can you address this?

adisbladis:
Fixed and applied in https://github.com/nix-community/poetry2nix/pull/471/commits/9ce2f75ff4063683bfd3f99a2aa4a88e824cfa1e.

PR #441, opened 2021-11-18
chore(pkgs/poetry): 1.1.11 -> 1.2.0a2

lovesegfault:


pkgs/poetry/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "poetry"
-version = "1.1.11"
+version = "1.2.0a2"
```
l0b0, line 3: Why upgrade to a [testing release](https://python-poetry.org/blog/announcing-poetry-1.2.0a2/)?

pkgs/poetry/update
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env nix-shell
 #! nix-shell -i bash -p curl nix-prefetch-github jq
 
-rev=$(curl -s https://api.github.com/repos/python-poetry/poetry/releases/latest | jq -r '.name')
+rev=1.2.0a2
```
l0b0, line 4: Doesn't this defeat the purpose of this script?

lovesegfault: I shouldn't have added that to the commit, it was a oneshot thing.

pkgs/poetry/pyproject.toml
lovesegfault: It's not meant to be merged now, just so that we can start playing with it

PR #447, opened 2021-11-22
overrides: add paramiko

rissson:


rissson:
That was fast, I couldn't even merge my PR on the project in which I needed this. I can just update poetry2nix there now. Thanks!

PR #448, opened 2021-11-22
overrides: add kerberos

g00pix:


overrides.nix
```diff
@@ -642,6 +642,9 @@ self: super:
     }
   );
 
+  kerberos = super.kerberos.overrideAttrs (old: {
+    nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ pkgs.libkrb5 ];
```
adisbladis, line 646: Are you sure this should be a native build input?
It looks to me like something that is just a regular build input, but I'm no expert in these libraries.

g00pix: I'm still having issues to understand the differences but all I can say is that it works for me if it's a native build input and it doesn't if it's a regular one.

adisbladis: Native vs non-native has to do with cross compilation.

> all I can say is that it works for me if it's a native build input and it doesn't if it's a regular one.

Ok, that's good enough for me :)

PR #452, opened 2021-11-29
Include the trailing slash in legacy index URL

gmacon:
The [PyPI Legacy API documentation][1] says that the distribution downloads for project can be accessed via a request

    GET /simple/<project>/

but the current implementation instead makes a request

    GET /simple/<project>

There exist some custom repository servers that give erroneous responses to requests without the slash, so include the slash.

[1]: https://warehouse.pypa.io/api-reference/legacy.html


adisbladis:
Thanks!

PR #458, opened 2021-12-08
Perform dead code elimination using deadnix

adisbladis:
deadnix: https://github.com/astro/deadnix


adisbladis:
Closing this as `deadnix` isn't even in nixpkgs yet, and will probably change.

PR #463, opened 2021-12-15
overrides: cryptography: fix cargo hash for versions >= 3.6.1

rissson:


rissson:
Thinking about this, it might be better to introduce a set with known versions and hashes and then iterate on that.

PR #464, opened 2021-12-16
overrides: matplotlib: use MPLSETUPCFG

veprbl:
Hi! Friendly upstream here. We made some change[1] that will make your `postPatch` override ineffective. This should make a backwards-compatible change to correct that. Also, I'm not a poetry user, so this is untested. I hope you can help me with that.

cc @K900

[1] https://github.com/NixOS/nixpkgs/commit/7dc44307515de130daf661814891d8844fdf701f

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -779,7 +779,7 @@ self: super:
         self.setuptools-scm-git-archive
       ];
 
-      MPLSETUPCFG = writeText "mplsetup.cfg" ''
+      MPLSETUPCFG = pkgs.writeText "mplsetup.cfg" ''
         [libs]
         system_freetype = True
         system_qhull = True
--- a/overrides.nix
+++ b/overrides.nix
@@ -779,18 +779,13 @@ self: super:
         self.setuptools-scm-git-archive
       ];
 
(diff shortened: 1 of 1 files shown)


overrides.nix
```diff
@@ -779,18 +779,13 @@ self: super:
         self.setuptools-scm-git-archive
       ];
 
-      postPatch = ''
-        cat > '' +
-      (if lib.versionAtLeast super.matplotlib.version "3.5.0" then "mplsetup.cfg" else "setup.cfg") +
```
adisbladis, line 784: We need to retain this if statement to be compatible with older versions of `matplotlib`, no?

adisbladis (changes requested):

overrides.nix
veprbl: I haven't tested it, but the MPLSETUPCFG was introduced 11 years ago https://github.com/matplotlib/matplotlib/commit/6713af0624c70ac1d0c1bc42461325f1e5f80363

adisbladis: @veprbl Thanks for the reference!

PR #466, opened 2021-12-19
build: remove bootstrap dependency on poetry

cpcloud:
This PR removes a dependency on `poetry` that gets put into most dependencies
as a `buildInputs` if poetry2nix isn't bootstrapping, which if I'm reading the
code correctly is nearly everywhere (`__isBootstrap` is `false` by default).

This is a problem when building using `crossSystem` (I'm trying to
build a Docker image using musl) because any dependency that overlapped with
`poetry`'s dependencies would cause both the build *and* host platform versions
of poetry's deps to be installed, causing a conflict during build.

It also appears that this bootstrap dependency is not necessary, at least
according to the current test suite. The only package that needed modification
was `pendulum`.

### DIFF
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -129,7 +129,6 @@ pythonPackages.callPackage
         ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) pythonPackages.setuptools
         ++ lib.optional (!isSource) (getManyLinuxDeps fileInfo.name).pkg
         ++ lib.optional isDirectory buildSystemPkgs
-        ++ lib.optional (!__isBootstrap) pythonPackages.poetry
       );
 
       propagatedBuildInputs =
--- a/overrides.nix
+++ b/overrides.nix
@@ -2103,7 +2103,7 @@ self: super:
   );
 
   pendulum = super.pendulum.overridePythonAttrs (old: {
-    buildInputs = (old.buildInputs or []) ++ [ self.poetry ];
+    buildInputs = (old.buildInputs or [ ]) ++ [ self.poetry ];
     # Technically incorrect, but fixes the build error..
     preInstall = lib.optionalString stdenv.isLinux ''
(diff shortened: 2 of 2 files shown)


cpcloud:
Here's a CI run where I'm building a musl-based image: https://github.com/cpcloud/protoletariat/runs/4576549106?check_suite_focus=true

cpcloud:
@adisbladis Thoughts on this PR?

mk-poetry-dep.nix
```diff
@@ -129,7 +129,6 @@ pythonPackages.callPackage
         ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) pythonPackages.setuptools
         ++ lib.optional (!isSource) (getManyLinuxDeps fileInfo.name).pkg
         ++ lib.optional isDirectory buildSystemPkgs
-        ++ lib.optional (!__isBootstrap) pythonPackages.poetry
```
adisbladis, line 132: This will break some downstreams (not all of the cases where this is required is covered by tests).
I would be open to something like `!__isBootstrap && !isCross` so that we can solve your issue while not breaking various projects using this code.

adisbladis (changes requested):

mk-poetry-dep.nix
cpcloud: Can you write down such a scenario? I'm happy to write one or more test cases, so that folks don't spend a ton of time tracking down the problem to this (like I did) without a test breaking.

adisbladis: This is really the same as any other build-system missing caused by https://github.com/python-poetry/poetry/issues/2789.

Adding `poetry` to each build was just a hack to circumvent that breakage for stuff depending on Poetry which I figured would be a fairly cheap workaround considering you are very likely to have poetry in your dependency graph anyway if you use this tool.

Maybe as a minimal reproduction case you could try an environment with https://github.com/tweag/pynixutil in it?

adisbladis:
I've decided to merge this since the only downside is that we need a bit more overrides in the stdlib, but it's not possible to fix the cross use case via overrides.

PR #480, opened 2021-12-30
ci: refactor

cpcloud:
- ci: use jq to construct builds matrix
- ci: bump to install cachix action v16


adisbladis:
Sweet, thanks!

PR #491, opened 2022-01-06
Adds overrides for humanize, inquirer

averagechris:
I'm not sure if either of these are maintainable, so any feedback is welcome here.

Particularly, I know that the requirements.txt for inquirer cannot be merged in as is. But I'm not sure what a better approach is.

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -534,6 +534,12 @@ self: super:
     propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ self.pyparsing ];
   });
 
+  humanize = super.humanize.overridePythonAttrs (old: {
+    postPatch = (old.postPatch or "") + ''
+      sed -i '/\[metadata\]/aversion = ${old.version}' setup.cfg
+    '';
+  });
+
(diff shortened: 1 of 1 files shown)


overrides.nix
```diff
@@ -619,6 +625,20 @@ self: super:
...
+      '';
+      preConfigure = (old.preConfigure or "") + ''cat << EOF > requirements.txt
+blessed==1.19.0
+readchar==2.0.1
+python-editor==1.0.4
+EOF
+      '';
+    }
+  );
+
```
averagechris, line 641: What's a better way to add this file. inquirer's setup.py calls a function that reads the requirements.txt file.

perhaps add a list of dependencies manually is better, avoiding requirements.txt altogether?

averagechris:
Actually it looks like inquirer recently updating their python packaging to be more standard as of https://github.com/magmax/python-inquirer/tree/v2.8.0 (~10 days ago)

I think I'm gonna close this PR instead of merging it since this override isn't needed with current versions. Hopefully anyone who needs the older version can find this PR via web search instead :D

PR #495, opened 2022-01-10
overrides.pyqt5: Set QT_PLUGIN_PATH and use wrapQtAppsHook

### LINKED ISSUE #283: pyqt cannot load plugins
This repository has a test that pyqt5 builds, but it does not test that it actually works. In reality, even a “hello world”, i.e. trying to show an empty window, fails with:

```text
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

This is probably somehow related to NixOS/nixpkgs#65399 and the solution is to set `$QT_PLUGIN_PATH` correctly. I did this by adding:

```nix
{
  postInstall = ''
    for i in "$out"/bin/*; do
      wrapProgram "$i" --prefix QT_PLUGIN_PATH : "${pkgs.qt5.qtbase}/${pkgs.qt5.qtbase.qtPluginPrefix}"
    done
  '';
}
```

Ideally, we’d want to make this happen automatically, and also somehow reuse `wrapQtAppsHook`. The problem with this hook is that it is too smart and it only wraps ELF binaries.

adisbladis:
Closes #283

### DIFF
--- a/overrides.nix
+++ b/overrides.nix
@@ -1382,6 +1382,7 @@ self: super:
           pkgs.qt5.qtwebchannel
           # self.pyqt5-sip
           self.sip
+          pkgs.qt5.wrapQtAppsHook
         ]
           ++ lib.optional withConnectivity pkgs.qt5.qtconnectivity
           ++ lib.optional withMultimedia pkgs.qt5.qtmultimedia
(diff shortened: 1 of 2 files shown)


adisbladis:
I'm not interested enough to push this through...

PR #503, opened 2022-01-12
Remove rec from template

Yajo:
It seems this is an anti-pattern, according to https://nix.dev/anti-patterns/language#rec-expression.

So, I suggest to remove it from there. Better not ship anti-patterns on templates, isn't it?

### DIFF
--- a/templates/app/flake.nix
+++ b/templates/app/flake.nix
@@ -24,11 +24,11 @@
           overlays = [ self.overlay ];
         };
       in
-      rec {
+      {
         apps = {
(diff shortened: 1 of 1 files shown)


adisbladis:
I totally agree and I hate `rec` with a passion!

Could you run `nixpkgs-fmt` and force-push this branch?

Yajo:
El mar, 11 de ene de 2022 a las 23:32:57 PM, adisbladis 
***@***.***> escribió:
> Could you run nixpkgs-fmt and force-push this branch?
> 

Done

PR #508, opened 2022-01-14
 overrides.pymssql: add override

lunik1:
Sorry for the PR spam, I would have bundled this with the pyrfr override but I forgot about it.


adisbladis:
No worries about the number of PRs :)

PR #512, opened 2022-01-17
 fetchGit: work-around bug in Nix 2.4

adisbladis:
This is a revival of #424 since it seems like upstream has zero interest in fixing this issue.


vale981:
is this bug still present?
i am getting 
```
Cannot find Git revision '404d51d295879edd93fb71c09373ef3273eefda6' in ref 'HEAD' of repository 'ssh://git@github.com/vale981/stocproc.git'! Please make sure that the rev exists on the ref you've specified or add allRefs = true; to fetchGit.
```

PR #513, opened 2022-01-17
Remove dontPreferSetupPy & don't remove pyproject.toml

adisbladis:
It seems like this bandaid to avoid some build-system related overrides causes more problems than it fixes.
Let's just deal with reality up-front and hope that https://github.com/python-poetry/poetry/issues/2789 will be fixed at some point.

Additionally this results in _much_ better error messages since you will be told exactly what input is missing.
With the faulty version detection you are simply told "the version is none".

### DIFF
--- a/hooks/pip-build-hook.sh
+++ b/hooks/pip-build-hook.sh
@@ -5,14 +5,6 @@ pipBuildPhase() {
     echo "Executing pipBuildPhase"
     runHook preBuild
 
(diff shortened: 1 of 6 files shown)


adisbladis:
This is somewhat related to https://github.com/nix-community/poetry2nix/pull/466 which I think I have changed my opinion on and will tackle soon.

PR #521, opened 2022-01-19
Add some overrides

K900:
Split into two commits to kinda keep related stuff together.


adisbladis:
Some additional test cases would be good I think. But let's not block this PR on tests. LGTM!

PR #522, opened 2022-01-19
Extended cross tests

adisbladis:


adisbladis:
Notably `cryptography` still fails because of some interaction with `cffi` that I don't know how to solve yet.
I'm going to leave it up to someone else to take up the mantle on continued cross efforts.

PR #531, opened 2022-01-22
Skip build jobs in CI if already cached

adisbladis:
Waiting times right now are pretty high and most jobs are cache hits anyway.
Let's skip jobs that are cached and make the Github scheduler aware of that.


adisbladis:
Github actions simple aren't very expressive... I don't know how to properly express this within the confines of their shitty yaml crap.

Issue #323, opened 2021-06-16
Can't install mypy 0.902


zupo:
I get the following error when trying to install mypy 0.902:

```
Processing /private/tmp/nix-build-python3.8-mypy-0.902.drv-0/mypy-0.902
    ERROR: Command errored out with exit status 1:
     command: /nix/store/3dvkfi06mzgszfwkhlrjnww1fj02ymbf-python3-3.8.9/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/tmp/nix-build-python3.8-mypy-0.902.drv-0/pip-req-build-8zacv0o5/setup.py'"'"'; __file__='"'"'/private/tmp/nix-build-python3.8-mypy-0.902.drv-0/pip-req-build-8zacv0o5/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/tmp/nix-build-python3.8-mypy-0.902.drv-0/pip-pip-egg-info-s0q3yqyp
         cwd: /private/tmp/nix-build-python3.8-mypy-0.902.drv-0/pip-req-build-8zacv0o5/
    Complete output (84 lines):
    mypy/modulefinder.py:432: error: Library stubs not installed for "toml" (or incompatible with Python 3.8)
    mypy/config_parser.py:10: error: Library stubs not installed for "toml" (or incompatible with Python 3.8)
    mypy/config_parser.py:10: note: Hint: "python3 -m pip install types-toml"
    mypy/config_parser.py:10: note: (or run "mypy --install-types" to install all missing stub packages)
    mypy/config_parser.py:10: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
    mypy/fastparse.py:109: error: Library stubs not installed for "typed_ast" (or incompatible with Python 3.8)
    mypy/fastparse.py:109: note: Hint: "python3 -m pip install types-typed-ast"
    mypy/fastparse2.py:57: error: Library stubs not installed for "typed_ast.ast27" (or incompatible with Python 3.8)
    mypy/fastparse2.py:68: error: Library stubs not installed for "typed_ast" (or incompatible with Python 3.8)
...
```

Because I saw `Hint: "python3 -m pip install types-toml"` in the error log, I tried to work around it with the following:

[...]

PR #534, opened 2022-01-25
Compile mypy once again

takeda:
So last time we disabled compilation, because it was broken since release of 0.900: https://github.com/nix-community/poetry2nix/issues/323

Looks like things blocking it got resolved and it could be enabled once again. This will increase build time, but then produce a much faster version of mypy.

Compilation could be disabled once again by overriding:
```
MYPY_USE_MYPYC=0
```

or time spent could be reduced by:
```
MYPYC_OPT_LEVEL=<value between 0 and 3>
```

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -973,11 +973,25 @@ lib.composeManyExtensions [
 
       mypy = super.mypy.overridePythonAttrs (
         old: {
-          MYPY_USE_MYPYC =
-            # is64bit: unfortunately the build would exhaust all possible memory on i686-linux.
-            stdenv.buildPlatform.is64bit
-            # Derivation fails to build since v0.900 if mypyc is enabled.
-            && lib.strings.versionOlder old.version "0.900";
+          patches = (old.patches or [ ]) ++ lib.optionals (lib.strings.versionAtLeast old.version "0.900") [
+            # FIXME: Remove patch after upstream has decided the proper solution.
+            #        https://github.com/python/mypy/pull/11143
+            (pkgs.fetchpatch {
+              url = "https://github.com/python/mypy/commit/f1755259d54330cd087cae763cd5bbbff26e3e8a.patch";
+              sha256 = "sha256-5gPahX2X6+/qUaqDQIGJGvh9lQ2EDtks2cpQutgbOHk=";
+            })
+          ];
+          buildInputs = (old.buildInputs or [ ]) ++ [
+            self.types-typed-ast
+          ];
+          # Compile mypy with mypyc, which makes mypy about 4 times faster. The compiled
+          # version is also the default in the wheels on Pypi that include binaries.
+          # is64bit: unfortunately the build would exhaust all possible memory on i686-linux.
+          MYPY_USE_MYPYC = stdenv.buildPlatform.is64bit;
+
+          # when testing reduce optimisation level to drastically reduce build time
+          # (default is 3)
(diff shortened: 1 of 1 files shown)


overrides/default.nix
```diff
@@ -973,11 +973,25 @@ lib.composeManyExtensions [
...
         old: {
-          MYPY_USE_MYPYC =
-            # is64bit: unfortunately the build would exhaust all possible memory on i686-linux.
-            stdenv.buildPlatform.is64bit
-            # Derivation fails to build since v0.900 if mypyc is enabled.
-            && lib.strings.versionOlder old.version "0.900";
+          patches = (old.patches or [ ]) ++ [
+            # FIXME: Remove patch after upstream has decided the proper solution.
+            #        https://github.com/python/mypy/pull/11143
+            (pkgs.fetchpatch {
```
adisbladis, line 979: This patch needs a version conditional.

overrides/default.nix
```diff
@@ -973,11 +973,25 @@ lib.composeManyExtensions [
...
+              url = "https://github.com/python/mypy/commit/f1755259d54330cd087cae763cd5bbbff26e3e8a.patch";
+              sha256 = "sha256-5gPahX2X6+/qUaqDQIGJGvh9lQ2EDtks2cpQutgbOHk=";
+            })
+          ];
+          buildInputs = old.buildInputs ++ [
```
adisbladis, line 984: The safe way to write this would be `old.buildInputs or [ ] ++ [`.

overrides/default.nix
```diff
@@ -973,11 +973,25 @@ lib.composeManyExtensions [
...
+            self.types-typed-ast
+          ];
+          # Compile mypy with mypyc, which makes mypy about 4 times faster. The compiled
+          # version is also the default in the wheels on Pypi that include binaries.
+          # is64bit: unfortunately the build would exhaust all possible memory on i686-linux.
+          MYPY_USE_MYPYC = stdenv.buildPlatform.is64bit;
```
adisbladis, line 990: Given our previous issues this should probably also have a version conditional.

adisbladis (changes requested):

overrides/default.nix
takeda: While I understand the other suggestions, I'm a bit confused about this one. The mypy was able to be compiled before 0.900 and this change allows it to be compiled after 0.900, so I'm not sure what versions I should use.

overrides/default.nix
takeda: Actually now as I looked into it more this patch is likely not needed for compilation itself.

It was added because the fix was done when adding mypy-zope package. This patch though fixes a bug that manifests especially in Nix, where we use large PYTHONPATH instead a single site-packages directory.

Should I remove it then?

adisbladis: Yes, please remove it if it's not required.

overrides/default.nix
adisbladis: Alright, I guess I was confused about the state of this issue.

overrides/default.nix
takeda: fixed

overrides/default.nix
takeda: fixed

overrides/default.nix
takeda: looks like that is necessary; added a version conditional

PR #544, opened 2022-02-02
More various fixes

zupo:
A few more fixes I found when working on a new project.

Again tested in our CI on Linux and macOS.

PR #551, opened 2022-02-05
Add hook to remove normal files from site-packages

adisbladis:
This is a fairly common packaging mistake that it's better that we solve in a hook than in one-by-one in overrides.


adisbladis:
This is somewhat related to #544

zupo (approved):
Looks good!

PR #558, opened 2022-02-16
overrides.entrypoints: add override

lunik1:
entrypoints needs flit_core to build

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -262,7 +262,7 @@
     "poetry-core"
   ],
   "entrypoints": [
-    "flit"
+    "flit-core"
   ],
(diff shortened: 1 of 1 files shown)


overrides/default.nix
```diff
@@ -407,6 +407,12 @@ lib.composeManyExtensions [
         }
       );
 
+      entrypoints = super.entrypoints.overridePythonAttrs (
```
adisbladis, line 410: Could you move this to the `build-systems.json` file?

adisbladis (changes requested):

lunik1:
My mistake, this already exists in `build-systems.json`

PR #559, opened 2022-02-21
overrides.entrypoints: entrypoints only requires flit-core

lunik1:
Following from #558, entrypoints only requires flit-core to build.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -262,7 +262,7 @@
     "poetry-core"
   ],
   "entrypoints": [
-    "flit"
+    "flit-core"
   ],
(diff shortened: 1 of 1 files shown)


adisbladis:
Are you sure that an older version of entrypoints doesn't require full flit for building?

lunik1:
Ah yes. It seems, 0.3 and earlier require `flit`.

adisbladis:
Is the addition of the full flit causing issues for you or was it an attempt at closure size reduction?

lunik1:
Just closure size reduction

adisbladis:
👌

PR #577, opened 2022-03-28
Add PEP 600 (new manylinux) support

### LINKED ISSUE #449: Support for wheels using PEP 600 manylinux platform tags
I've noticed that the `manylinux1`, `manylinux2010`, and `manylinux2014` platform tags are the only ones supported here. A new style of tag, defined in [PEP 600](https://www.python.org/dev/peps/pep-0600/) now exists using the pattern `manylinux_${GLIBCMAJOR}_${GLIBCMINOR}_${ARCH}`. PEP 600 also redefines all of the existing tags as aliases of tags in the new style.

I think this was the root cause of #306.

I think it would be good to support this new style of tag, though I don't have an immediate suggestion as to how that would best be done.

SemMulder:
This adds support for [PEP 600](https://peps.python.org/pep-0600/), which covers a new tagging system for manylinux. Merging this would close #449.

I added a test that installs [open3d](https://pypi.org/project/open3d/) in `pep600.nix`.

While implementing this I also did the following to get this test to pass:
- did a minor cleanup of `lib.strings` references in `pep425.nix`,
- Added support for multiple (dot-separated) platform tags:
  - (e.g. `numpy-1.22.3-pp38-pypy38_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`)
- Fixed a minor bug where the build system was added to `nativeBuildInputs` in overrides.nix for wheels.
- Added some overrides for `open3d` and `jupyter-packaging`.

Since this has turned into quite a lot of stuff, I suggest reviewing this PR per commit.

Since this is my first time contributing to poetry2nix, please let me know if this PR fulfills the requirements. I'm happy to change stuff around to get this merged :).

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -79,6 +79,7 @@ let
     if lib.strings.hasInfix "manylinux1" f then { pkg = [ ml.manylinux1 ]; str = "1"; }
     else if lib.strings.hasInfix "manylinux2010" f then { pkg = [ ml.manylinux2010 ]; str = "2010"; }
     else if lib.strings.hasInfix "manylinux2014" f then { pkg = [ ml.manylinux2014 ]; str = "2014"; }
+    else if lib.strings.hasInfix "manylinux_" f then { pkg = [ ml.manylinux2014 ]; str = "pep600"; }
     else { pkg = [ ]; str = null; };
 
   # Predict URL from the PyPI index.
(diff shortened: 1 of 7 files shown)


SemMulder:
Disabled the test for PEP 600 on MacOS due to open3d not being available on Mac. Let me know if you want me to change to another package (I had trouble finding a simpler package, but I can dig deeper ;)).

SemMulder:
The tests failed due to a spurious network error, but I can't seem to restart the failed jobs.

adisbladis (approved):
Barring any test failures this looks very good to me!

PR #567, opened 2022-03-08
pin dependencies of the update script

teto:
right now the pkgs/poetry/update script looks for dependencies in the user <nixpkgs>.
I made it so that it sets NIX_PATH=nixpkgs=theOneFromTheFlake

My initial motivation was to bump poetry since I hit https://github.com/python-poetry/poetry/issues/4515
I haven't done it yet because I hit the following issue I am not sure how to address
```
poetry2nix/pkgs/poetry on  pin-nixpkgs via 🐍 took 26s 
➜ ./update
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels/nixos' does not exist, ignoring
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «string»:1:9:

            1| (import <nixpkgs> {}).bashInteractive
             |         ^
will use bash from your environment
+ echo

++ curl -s https://api.github.com/repos/python-poetry/poetry/releases/latest
++ jq -r .name
+ rev=1.1.13
++ nix eval -f nixpkgs.nix --raw
+ export NIX_PATH=nixpkgs=/nix/store/fg5cspmfnm9qpi57lgbprspnjhj73y2a-source
+ NIX_PATH=nixpkgs=/nix/store/fg5cspmfnm9qpi57lgbprspnjhj73y2a-source
+ nix-prefetch-github --rev 1.1.13 python-poetry poetry
+ echo
++ nix-build --no-out-link --expr 'with import (import ./nixpkgs.nix ) {}; fetchFromGitHub (lib.importJSON ./src.json)' --show-trace
+ src=/nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source
+ echo 'Printing source'
Printing source
+ echo /nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source
/nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source
+ cp /nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source/pyproject.toml /nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source/poetry.lock .
+ nix-shell -p poetry --run 'poetry lock'
Updating dependencies
Resolving dependencies... (211.6s)

  SolverProblemError

  The current project's Python requirement (>=2.7,<2.8 || >=3.5,<4.0) is not compatible with some of the required packages Python requirement:
[...]

PR #580, opened 2022-03-31
Remove flake.lock and gitignore it

adisbladis:
Poetry2nix doesn't use flakes for development and the lock is rarely updated.

Related PR #567


takeda:
@adisbladis today I had a weird issue where I refreshed poetry2nix yet I was still getting old version (the one that did not include your latest cryptography package changes).

I eventually managed to get newer version after some removing and placing things back. I'm wondering if when flake.lock is removed `nix flake update` skipped the update.

I use flake in my project, and old nixpkgs is not an issue for me as I have:
```nix
poetry2nix.inputs.nixpkgs.follows = "nixpkgs";
```

I also noticed that you mentioned that you're not too happy about flakes. I'm wondering what the issue is. I started using them recently and so far I like them, but perhaps I haven't used them fully yet. Is the problem with rigidity of outputs?

PR #588, opened 2022-04-12
build systems

cpcloud:
- build-systems.json: add poetry-dynamic-versioning
- build-systems.json: add pyparsing


overrides/build-systems.json
```diff
@@ -810,6 +810,9 @@
   "poetry2conda": [
     "poetry"
   ],
+  "poetry-dyanmic-versioning": [
```
takeda, line 813: this looks like a typo

PR #589, opened 2022-04-14
fix: avoid infinite recursion while searching for path #585

adisbladis:
Since I'm not allowed to force push #585 I'm opening this PR for a CI run before merge.

Closes #585


Yajo:
In #585 I was fixing the problem while using nix flakes, but it still existed while using the stable `nix-build` (just found out that recently).

This fixes both usage scenarios though. Thank you very much!

PR #614, opened 2022-05-07
overrides: adding python-trovo and python-twitter

WolfangAukang:


overrides/default.nix
```diff
@@ -1758,6 +1758,14 @@ lib.composeManyExtensions [
         }
       );
 
+      python-trovo = super.python-trovo.overridePythonAttrs (old: {
+        buildInputs = (old.buildInputs or [ ]) ++ [ self.poetry ];
```
adisbladis, line 1762: Could you add this one to `build-systems.json` instead?

WolfangAukang: Sure thing

adisbladis: Thanks :)
It's so that one day in the future we could potentially get all build-system's from poetry.lock and drop all our custom overrides.

PR #617, opened 2022-05-08
Don't change format of pyproject.toml when removing special dependencies

### LINKED ISSUE #613: remove-special-dependencies.sh hook reformats pyproject.toml, breaking Scipy.
### Describe the issue

The `remove-*-dependencies.sh` hooks reformat `pyproject.toml` without indentation. This breaks a [(pretty bad) parser in `scipy/tools/cythonize.py`](https://github.com/scipy/scipy/blob/3595b3947faacbedf76de4cb410e0a4210059117/tools/cythonize.py#L64). Which, in turn, breaks the build.

Adding a `-i` (i.e. indent) flag in [remove-special-dependencies.sh](https://github.com/nix-community/poetry2nix/blob/9927a08fd9537e86a7769b00ab7c231066ae1fb3/hooks/remove-special-dependencies.sh#L13) will most likely solve that.

Another approach could be to leave the pyproject.toml untouched if there are no special dependencies in it.

@cpcloud which approach would you prefer? (Assuming we want to fix this :wink:).

SemMulder:
This is needed because Scipy parses pyproject.toml by hand, under
the assumption of one dependency per line.

This closes #613.

### DIFF
--- a/hooks/pyproject-without-special-deps.py
+++ b/hooks/pyproject-without-special-deps.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python
-# Patch out special dependencies (git and path) from a pyproject.json file
+# Patch out special dependencies (git and path) from a pyproject.toml file
 
 import argparse
-import json
 import sys
+
+import tomlkit
 
 
 def main(input, output, fields_to_remove):
-    data = json.load(input)
+    data = tomlkit.loads(input.read())
 
     try:
         deps = data["tool"]["poetry"]["dependencies"]
@@ -22,12 +23,7 @@ def main(input, output, fields_to_remove):
                 if any_removed:
                     dep["version"] = "*"
 
-    # Set ensure_ascii to False because TOML is valid UTF-8 so text that can't
-    # be represented in ASCII is perfectly legitimate
-    # HACK: Setting ensure_asscii to False breaks Python2 for some dependencies (like cachy==0.3.0)
-    json.dump(
-        data, output, separators=(",", ":"), ensure_ascii=sys.version_info.major < 3
-    )
+    output.write(tomlkit.dumps(data))
 
 
 if __name__ == "__main__":
@@ -37,20 +33,20 @@ if __name__ == "__main__":
         "--input",
         type=argparse.FileType("r"),
         default=sys.stdin,
-        help="Location from which to read input JSON",
+        help="Location from which to read input TOML",
     )
     p.add_argument(
         "-o",
         "--output",
         type=argparse.FileType("w"),
         default=sys.stdout,
-        help="Location to write output JSON",
+        help="Location to write output TOML",
     )
     p.add_argument(
         "-f",
         "--fields-to-remove",
         nargs="+",
-        help="The fields to remove from the dependency's JSON",
+        help="The fields to remove from the dependency's TOML",
     )
 
     args = p.parse_args()
--- a/hooks/default.nix
+++ b/hooks/default.nix
@@ -3,15 +3,18 @@
 , makeSetupHook
 , wheel
 , pip
+, pkgs
 }:
 let
   callPackage = python.pythonForBuild.pkgs.callPackage;
   pythonInterpreter = python.pythonForBuild.interpreter;
   pythonSitePackages = python.sitePackages;
+  nonOverlayedPython = pkgs.python3.pythonForBuild.withPackages (ps: [ ps.tomlkit ]);
 in
 {
-
-  removePathDependenciesHook = callPackage
+  # NOTE: We have to use a non-overlayed Python here because otherwise we run into an infinite recursion
+  # because building of tomlkit and its dependencies also use these hooks.
+  removePathDependenciesHook = nonOverlayedPython.pkgs.callPackage
     (
       {}:
       makeSetupHook
@@ -19,8 +22,7 @@ in
           name = "remove-path-dependencies.sh";
           deps = [ ];
           substitutions = {
-            inherit pythonInterpreter;
-            yj = "${buildPackages.yj}/bin/yj";
+            pythonInterpreter = nonOverlayedPython.interpreter;
             pyprojectPatchScript = "${./pyproject-without-special-deps.py}";
             fields = [ "path" ];
             kind = "path";
@@ -29,15 +31,15 @@ in
     )
     { };
 
-  removeGitDependenciesHook = callPackage
-    ({}:
+  removeGitDependenciesHook = nonOverlayedPython.pkgs.callPackage
+    (
+      {}:
       makeSetupHook
         {
           name = "remove-git-dependencies.sh";
           deps = [ ];
           substitutions = {
-            inherit pythonInterpreter;
-            yj = "${buildPackages.yj}/bin/yj";
+            pythonInterpreter = nonOverlayedPython.interpreter;
             pyprojectPatchScript = "${./pyproject-without-special-deps.py}";
             fields = [ "git" "branch" "rev" "tag" ];
             kind = "git";
@@ -89,6 +91,4 @@ in
         } ./wheel-unpack-hook.sh
     )
     { };
-
-
 }
--- a/hooks/default.nix
+++ b/hooks/default.nix
@@ -9,44 +9,40 @@ let
   callPackage = python.pythonForBuild.pkgs.callPackage;
   pythonInterpreter = python.pythonForBuild.interpreter;
   pythonSitePackages = python.sitePackages;
+
   nonOverlayedPython = pkgs.python3.pythonForBuild.withPackages (ps: [ ps.tomlkit ]);
+  makeRemoveSpecialDependenciesHook = { fields, kind }:
+    nonOverlayedPython.pkgs.callPackage
+      (
+        {}:
+        makeSetupHook
+          {
+            name = "remove-path-dependencies.sh";
+            deps = [ ];
+            substitutions = {
+              # NOTE: We have to use a non-overlayed Python here because otherwise we run into an infinite recursion
+              # because building of tomlkit and its dependencies also use these hooks.
+              pythonPath = nonOverlayedPython.pkgs.makePythonPath [ nonOverlayedPython ];
+              pythonInterpreter = nonOverlayedPython.interpreter;
+              pyprojectPatchScript = "${./pyproject-without-special-deps.py}";
+              fields = fields;
+              kind = kind;
+            };
+          } ./remove-special-dependencies.sh
+      )
+      { };
 in
 {
-  # NOTE: We have to use a non-overlayed Python here because otherwise we run into an infinite recursion
-  # because building of tomlkit and its dependencies also use these hooks.
-  removePathDependenciesHook = nonOverlayedPython.pkgs.callPackage
-    (
-      {}:
-      makeSetupHook
-        {
-          name = "remove-path-dependencies.sh";
-          deps = [ ];
-          substitutions = {
-            pythonInterpreter = nonOverlayedPython.interpreter;
-            pyprojectPatchScript = "${./pyproject-without-special-deps.py}";
-            fields = [ "path" ];
-            kind = "path";
-          };
-        } ./remove-special-dependencies.sh
-    )
-    { };
+  removePathDependenciesHook = makeRemoveSpecialDependenciesHook {
+    fields = [ "path" ];
+    kind = "path";
+  };
 
-  removeGitDependenciesHook = nonOverlayedPython.pkgs.callPackage
-    (
-      {}:
-      makeSetupHook
-        {
-          name = "remove-git-dependencies.sh";
-          deps = [ ];
-          substitutions = {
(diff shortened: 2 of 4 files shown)


hooks/pyproject-without-special-deps.py
```diff
@@ -1,13 +1,14 @@
...
 import argparse
-import json
 import sys
 
+import tomlkit
+
 
 def main(input, output, fields_to_remove):
-    data = json.load(input)
+    data = tomlkit.loads(input.read())
```
cpcloud, line 11: Would `tomlkit.load(input)` work?

hooks/pyproject-without-special-deps.py
```diff
@@ -22,12 +23,7 @@ def main(input, output, fields_to_remove):
                 if any_removed:
                     dep["version"] = "*"
 
-    # Set ensure_ascii to False because TOML is valid UTF-8 so text that can't
-    # be represented in ASCII is perfectly legitimate
-    # HACK: Setting ensure_asscii to False breaks Python2 for some dependencies (like cachy==0.3.0)
-    json.dump(
-        data, output, separators=(",", ":"), ensure_ascii=sys.version_info.major < 3
-    )
+    output.write(tomlkit.dumps(data))
```
cpcloud, line 26: Would `tomlkit.dump(data, output)` work?

SemMulder:
Just pushed a small change to fix the build. Specifying `tomlkit` as dependency in `callPackage` tries to install it twice. Probably something to do with the fact that `poetry-core` which is needed to build `tomlkit` also depends on `tomlkit`.

hooks/pyproject-without-special-deps.py
SemMulder: I think it will, let me check. I was trying to use a non-overlayed Python before, which then could be Python 2.7. In that case we got an ancient version of `tomlkit` that didn't have those functions yet.

SemMulder: But since I'm using an non-overlayed Python now, it should just work.

cpcloud (approved):
LGTM.

hooks/pyproject-without-special-deps.py
SemMulder: It doesn't :(

```
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
Sourcing pip-install-hook
Using pipInstallPhase
Using pythonImportsCheckPhase
unpacking sources
unpacking source archive /nix/store/h7bc0m662rcbci6x107allj8wkkkwa15-source
source root is source
setting SOURCE_DATE_EPOCH to timestamp 315619200 of file source/tox.ini
patching sources
Removing path dependencies
  File "/nix/store/6nqqz004095n2b5h55czcqvqidqngnv8-pyproject-without-special-deps.py", line 53, in <module>
    main(args.input, args.output, args.fields_to_remove)
  File "/nix/store/6nqqz004095n2b5h55czcqvqidqngnv8-pyproject-without-special-deps.py", line 11, in main
    data = tomlkit.load(input)
AttributeError: module 'tomlkit' has no attribute 'load'
```

SemMulder:
Looks like cachix is having issues.

SemMulder:
Hmm, somehow `enum34` is getting into the non-overlayed environment. [It shouldn't though](https://github.com/NixOS/nixpkgs/blob/47388bbd05ed056f4590999ee5b5f926857990b0/pkgs/development/python-modules/tomlkit/default.nix#L15). Not sure what is going on, will have to check later.

cpcloud:
One issue is that the last version of tomlkit that supports 2.7 is 0.7.2 (https://github.com/sdispater/tomlkit/releases/tag/0.8.0). I'll put up a PR to update nixpkgs upstream to remove the enum34 dependency, though this will probably have to go in staging, I suspect `tomlkit` has a large number of transitive dependencies.

SemMulder:
> One issue is that the last version of tomlkit that supports 2.7 is 0.7.2 

Hmm, I'm explicitly requesting Python 3, so that shouldn't matter right?

Looking at the traceback you can see that Python3.9 is being used:

```
 File "/nix/store/6azsi7klkz5b7jvha004dc6pbpdgar9x-pyproject-without-special-deps.py", line 4, in <module>
   import argparse
 File "/nix/store/33hlmb4afkk7l2fapfik80m6yknjxq10-python3-3.9.12/lib/python3.9/argparse.py", line 89, in <module>
   import re as _re
 File "/nix/store/33hlmb4afkk7l2fapfik80m6yknjxq10-python3-3.9.12/lib/python3.9/re.py", line 145, in <module>
   class RegexFlag(enum.IntFlag):
AttributeError: module 'enum' has no attribute 'IntFlag'
```

The strange thing is that it still seems like `enum34` is installed into this environment, where it shouldn't according to the definition in nixpkgs:

```
propagatedBuildInputs =
  lib.optionals isPy27 [ enum34 functools32 ]
  ++ lib.optional isPy27 typing;
```

cpcloud:
> > I suspect tomlkit has a large number of transitive dependencies.
> 
> It is not that bad I think. At least for recent versions there are no runtime dependencies, and for build-time only `poetry-core`.
> 
> [sdispater/tomlkit@`721acca`/pyproject.toml#L20](https://github.com/sdispater/tomlkit/blob/721acca52b39d515a5d04d66f1478277688ab5be/pyproject.toml#L20) [sdispater/tomlkit@`721acca`/pyproject.toml#L60](https://github.com/sdispater/tomlkit/blob/721acca52b39d515a5d04d66f1478277688ab5be/pyproject.toml#L60)

Ah, sorry I meant transitive **dependents**.

cpcloud:
@SemMulder Would the `-i` solution work?

SemMulder:
> @SemMulder Would the `-i` solution work?

Unfortunately not :disappointed:. I tried it, but it didn't put array elements on their own line.

SemMulder:
And as far as I could see there are no standalone TOML formatters.

SemMulder:
Maybe we can use [`nix why-depends`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-why-depends.html)? But I'll have to do it later, am at work currently.

SemMulder:
Oh, and the code as is does at least kind of work. I'm using it in one of my projects, and there it works fine. So it seems to be something specific to that test/environment.

cpcloud:
Looks like somehow `nixops`, but not `nixopsUnstable` introduces this problem.

SemMulder:
Maybe a `PYTHONPATH` variable is leaking through somewhere?

SemMulder:
That does seem to be the case:
```
~/git/poetry2nix (indent-pyproject-toml-after-removing-special-deps) » nix-shell /nix/store/lzv8afvf6kyghiysml61cckzhir1h9vc-python2.7-cryptography-3.3.2.drv --run 'env | grep PYTHONPATH'                                                                                                                     sem@semlaptop2
Sourcing pip-build-hook
Using pipShellHook
Sourcing pip-install-hook
Using pipInstallPhase
Using pythonImportsCheckPhase
Executing pipShellHook
Finished executing pipShellHook
PYTHONPATH=/nix/store/s5659c8ml4j51gkx13v6n2l3xmjhh8z3-python-2.7.18/lib/python2.7/site-packages:/nix/store/jz53r2890ivg97h85g68rgnsjbd35pfz-python2.7-setuptools-44.0.0/lib/python2.7/site-packages:/nix/store/jm3hf02ypz40r9nxy73nhmhr0m1rz45m-python2.7-pip-20.3.4/lib/python2.7/site-packages:/nix/store/p5mfy8h7q9w615bg8jr6ziq13i84nw2f-python2.7-wheel-0.37.1/lib/python2.7/site-packages:/nix/store/20n5841zc1vx86iqck4gq5dp9qsrr51n-python2.7-cffi-1.15.0/lib/python2.7/site-packages:/nix/store/dz9c0mf87ifd9wa5r0g0bcilchxccz4r-python2.7-pycparser-2.21/lib/python2.7/site-packages:/nix/store/0hg7633mj4fw2y8vhxrxa7q98wn2kqzh-python2.7-setuptools_scm-5.0.2/lib/python2.7/site-packages:/nix/store/ldri5am6q254md4scnx8q00z866lb713-python2.7-toml-0.10.2/lib/python2.7/site-packages:/nix/store/b9yhypg6ar851h99rvq74ixhi6xm9vzp-python2.7-enum34-1.1.10/lib/python2.7/site-packages:/nix/store/97l3ha0dp9286yqn4swca75isrmjpkzp-python2.7-ipaddress-1.0.23/lib/python2.7/site-packages:/nix/store/pwkk1nwbl226c25318j8c44ycbqhm9xd-python2.7-six-1.16.0/lib/python2.7/site-packages
```

A bit hard to read, but `/nix/store/b9yhypg6ar851h99rvq74ixhi6xm9vzp-python2.7-enum34-1.1.10/lib/python2.7/site-packages` is in there. I think that is just because of Python modules in `propagatedBuildInputs`?

We should be able to fix this by resetting the `PYTHONPATH` at the top of our hook script.

Give me a few minutes...

SemMulder:
That also explains why this happened:
> Just pushed a small change to fix the build. Specifying tomlkit as dependency in callPackage tries to install it twice. Probably something to do with the fact that poetry-core which is needed to build tomlkit also depends on tomlkit.

SemMulder:
Alright, that should fix it. I tested it locally, and now it works :smile:.

Good thing there are tests, this was quite subtle!

SemMulder:
Hard to say indeed :disappointed:. I get the exact same derivations locally:
```
~/git/poetry2nix (indent-pyproject-toml-after-removing-special-deps) » nix-build-uncached -build-flags "-L" --keep-going --show-trace tests/default.nix -A uwsgi
$ nix-build --dry-run --keep-going --show-trace tests/default.nix -A uwsgi
these 8 derivations will be built:
  /nix/store/fygn6sbbiqprxs7vwpr5nba750xpx301-uWSGI-2.0.19.1.tar.gz.drv
  /nix/store/s916g54n7dxxq710kbrb4m6p9d67n9wi-builder.pl.drv
  /nix/store/p75v53zkmbj3026r86d6s6cw2gr8560y-python3-3.9.12-env.drv
  /nix/store/imvy0hbnjjk8843yw0dhsb2pp8jsx9l4-remove-path-dependencies.sh.drv
  /nix/store/pr2y72vsz4y8mxv3rv9c9z42m7af00a4-remove-path-dependencies.sh.drv
  /nix/store/hvfqhqm3nb1s40r1gk5wj0k1fmfk5p7a-python3.9-uwsgi-2.0.19.1.drv
  /nix/store/z4h2z1w67maih0zpav2ipknx23dvfv0h-python3.9-poetry-1.1.13.drv
  /nix/store/1rnfzslz3d67dm3g440kkzd77lr8agn4-python3.9-test-uwsgi-0.1.0.drv
```
and here they build successfully. So I guess something is interfering on the agent, or, less likely, a bug in Nix?

I get the same warnings locally, but locally they don't break the build.

SemMulder:
Got something. Running `nix-build-uncached -build-flags "-L" -j 2 --cores 2 --keep-going --show-trace tests/default.nix -A uwsgi` does break the build.

So the issue apparently doesn't occur with 16 threads (what my laptop has). Interesting...

adisbladis:
> So the issue apparently doesn't occur with 16 threads (what my laptop has). Interesting...

I observed a test failure once out of ~15 attempts on my 16 thread laptop.
I think it must be something timing related that's just very unlikely to happen on larger thread counts..

SemMulder:
Yeah, some race condition or the like indeed.

I do get a higher failure rate (~75%) when running `nix-collect-garbage` before the run it seems. But not sure, it might not be, these kind of issues are tricky.

Any ideas on how to proceed?

adisbladis:
> Any ideas on how to proceed?

For now I've commented out that test since it's flaky and I think we've established with an extremely high level of certainty that the flakiness is unrelated to your change.

SemMulder:
> For now I've commented out that test since it's flaky and I think we've established with an extremely high level of certainty that the flakiness is unrelated to your change.

Thank you :smile:.

I'll try to get an `strace` of the error, but so far no luck in triggering it under `strace` of course :wink:.

adisbladis:
I just cancelled the CI run to implement some CI improvements in #622 that I will rebase this on, hopefully that will make mass rebuilds like this much more manageable.

SemMulder:
@adisbladis 

Regarding the race condition, I got an `strace` and got the following:

```
[thread 9][gcc] core/gateway.o
```
leads to
```
execve("/bin/sh", ["sh", "-c", "gcc -c -O2 -I. -Wall -D_LARGEFIL"...], 0x1499fd0 /* 152 vars */) = -1 EFAULT (Bad address)
```

Which probably means someone is not terminating the argument list in the call to `execve`. I'll see later if I can file a bug somewhere else...

SemMulder:
Seems like it might be a bug in Python. The failing call is a call to [os.system](https://github.com/unbit/uwsgi/blob/ca8bc6acc98b2c062b5b00668acd851e210fcbca/uwsgiconfig.py#L121).

PR #638, opened 2022-05-13
shapely: use correct python version for ast.unparse when using python3.8

cpcloud:
- shapely: add test
- shapely: use python3Minimal to ensure ast.unparse exists


cpcloud:
I can confirm the test I've added here fails without the `overrides/default.nix` change.

adisbladis (approved):

PR #643, opened 2022-05-18
traitlets switched build to hatch

takeda:
Two days ago traitlets switched to hatch: https://github.com/ipython/traitlets/pull/728 starting with version 5.2.1.


takeda:
@adisbladis I added it back

takeda:
So hatchling requires pluggy 1.0.0. This was breaking one of the test. I had some problems updating that dependency as poetry had problems installing pillow. I ended up doing minimal update to `common-pkgs-2`. There is a possibility that other tests might need to be fixed, we'll see.

takeda:
Ok, so this one I'm puzzled by.

While it is Pillow, this is different than the issue I described above (poetry did not download a wheel and instead pulled a tarball which it failed to compile).

I actually tested `common-pkgs-2` on mac and it built fine:
```
$ nix-build --keep-going --show-trace tests/default.nix -A common-pkgs-2
/nix/store/9zq2rlbai2xmxhpna550phxjr7a9y9kq-python3.9-common-pkgs-2-0.1.0
```

takeda:
@adisbladis I don't know how to resolve this issue. I called `nix-collect-garbage -d` on my Mac and `nix-build --keep-going --show-trace tests/default.nix -A common-pkgs-2` still succeeded there. I can't reproduce this error locally.

sigprof:
I encountered a similar problem building Pillow on GitHub CI. Apparently the build process finds the openjpeg library at `/usr/local/Cellar/openjpeg/2.5.0/include/openjpeg-2.5` and tries to use it, then fails later, because the library is not actually in the Nix store.

The Nixpkgs package for Pillow fixes the problem by patching out the guessing code from build scripts (see `preConfigure` in https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/pillow/generic.nix). It is possible to reuse the patching code in poetry2nix by adding an override like this:
```nix
      pillow = super.pillow.overridePythonAttrs(old: {
        # Use preConfigure from nixpkgs to fix library detection issues and
        # impurities which can break the build process; this also requires
        # adding propagatedBuildInputs and buildInputs from the same source.
        propagatedBuildInputs = (old.buildInputs or []) ++ pkgs.python3.pkgs.pillow.propagatedBuildInputs;
        buildInputs = (old.buildInputs or []) ++ pkgs.python3.pkgs.pillow.buildInputs;
        preConfigure = (old.preConfigure or "") + pkgs.python3.pkgs.pillow.preConfigure;
      });
```
However, I don't actually have a Mac and cannot test this on a real machine. For linux-x86_64, however, this change actually results in enabling most of options in Pillow except RAQM (and adding the appropriate dependencies for external libraries).

adisbladis:
This was fixed in #663.

Issue #656, opened 2022-06-07
mypy 0.960 patches fail with latest overrides


l0b0:
### Describe the issue

Using nixpkgs 22.05 with the [latest poetry2nix mypy overrides](https://github.com/nix-community/poetry2nix/blob/73fc21abbed96024f6533da259d5b01b1337ae4d/overrides/default.nix#L1087-L1120=) it fails to build:

```console
$ nix-shell 
these 2 derivations will be built:
  /nix/store/45q61q8r1ncql1xa3h193k2cqykkxamz-python3.8-mypy-0.960.drv
  /nix/store/zhfgc0fb7mrd0agw9vakfqxdb7jzsd8l-python3-3.8.13-env.drv
building '/nix/store/45q61q8r1ncql1xa3h193k2cqykkxamz-python3.8-mypy-0.960.drv'...
Sourcing python-remove-tests-dir-hook
Sourcing python-catch-conflicts-hook.sh
Sourcing python-remove-bin-bytecode-hook.sh
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
Sourcing pip-install-hook
Using pipInstallPhase
Sourcing python-imports-check-hook.sh
Using pythonImportsCheckPhase
Sourcing python-namespaces-hook
unpacking sources
unpacking source archive /nix/store/ij2br9yy0k1lkjb1n7d554wy9nxb5c5k-mypy-0.960.tar.gz
source root is mypy-0.960
setting SOURCE_DATE_EPOCH to timestamp 1653397510 of file mypy-0.960/setup.cfg
patching sources
applying patch /nix/store/g1vpfnjn785nq00qfpcmxjf8frzgxj65-e7869f05751561958b946b562093397027f6d5fa.patch
patching file mypy/main.py
Hunk #2 succeeded at 1043 (offset 10 lines).
patching file mypy/modulefinder.py
Hunk #1 FAILED at 30.
Hunk #2 succeeded at 724 (offset 116 lines).
Hunk #3 FAILED at 617.
Hunk #4 succeeded at 882 (offset 122 lines).
Hunk #5 succeeded at 892 (offset 122 lines).
2 out of 5 hunks FAILED -- saving rejects to file mypy/modulefinder.py.rej
patching file mypy/pyinfo.py
patching file mypy/test/testcmdline.py
patching file test-data/unit/cmdline.test
[...]

PR #657, opened 2022-06-10
Updated patch for mypy 0.960+

takeda:
The old patch URL did not apply cleanly for 0.960 and 0.961.

Looks like the PR was eventually merged so hopefully these might be the last versions needing a patch.

Fixes https://github.com/nix-community/poetry2nix/issues/656

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1112,8 +1112,8 @@ lib.composeManyExtensions [
             })
           ] ++ lib.optionals (lib.strings.versionAtLeast old.version "0.960") [
             (pkgs.fetchpatch {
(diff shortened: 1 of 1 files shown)


l0b0:
I can't seem to get the latest override to work with mypy 0.960 or 0.961 on nixpkgs stable 22.05. Do I have to use unstable?

takeda:
@l0b0 depends on the error, but likely the answer is yeah, a nixpkgs and packages in PyPI change locally have different dependencies.

 You can also override the build and define `MYPY_USE_MYPYC = false`. That should disable compiling of mypy. It will unfortunately be much slower, but also it will have far less issues and likely work out of the box.

PR #662, opened 2022-06-15
build-systems: add hatchling to traitlets

trepetti:
traitlets switched to the hatchling backend starting with [version 5.2.1](https://github.com/ipython/traitlets/releases/tag/5.2.1).

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -1690,7 +1690,8 @@
     "pbr"
   ],
   "traitlets": [
-    "flit-core"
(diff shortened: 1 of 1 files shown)


trepetti:
I wonder if it is possible to encode the version at which the build system switched over in `build-system.json`. It looks like going the multiple build system list approach without this information, some other packages that depend on pre-switch versions of the package in question fail to correctly resolve their existing dependencies due to taking on the new build system as a dependency unnecessarily.

Issue #655, opened 2022-06-07
How to set up build system override in `mkPoetryEnv`?


l0b0:
### Describe the issue

As a relative Nix newbie it's not obvious how to create a new override to emulate `addBuildSystem` (I'm trying to avoid changing `build-systems.json` itself, if that makes sense). For example, jsonschema 4.6.0 seems to require hatchling, but the override

```
jsonschema = super.jsonschema.overridePythonAttrs (
  old: {
    nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [
      self.hatchling
      self.hatch-vcs
    ];
  }
);
```

just results in a different error:

> ERROR: Could not find a version that satisfies the requirement pluggy>=1.0.0; python_version > "3" (from hatchling) (from versions: none)

Does this error require a different override, or did I do something wrong with my hatchling override?

sigprof:
The above override actually works for me if I use a recent nixpkgs snapshot that has usable `hatchling` and `hatch-vcs` Python packages.

Another option is to add the required modules to `pyproject.toml` (e.g., to the `[tool.poetry.dev-dependencies]` section), then poetry2nix would use the module versions specified there instead of taking whatever is in nixpkgs (and would also pick up any dependencies like that `pluggy` module). However, you may run into version conflicts with other modules; e.g., this is what I ended up with at the moment:
```toml
# These dependencies are required by the jsonschema >= 4.6.0 build system, but
# are not detected automatically; they are also not present in the used Nixpkgs
# snapshot, so need to be obtained through Poetry.
hatchling = "*"
hatch-vcs = "*"

# The `pytest` module in the used Nixpkgs snapshot has an upper bound on the
# `pluggy` dependency, which conflicts with the dependency of the `hatchling`
# module; upgrading the `pytest` module fixes the conflict.
pytest = "*"

# Building the `tomli` module, which is in the dependency tree of `hatchling`,
# requires a newer `flit` module than found in the used Nixpkgs snapshot.
flit = "*"
```

I suppose that the PR to add the override for `jsonschema` would need to use `if (lib.versionAtLeast super.jsonschema.version "4.6.0")` to keep older module versions working as before.

l0b0:
@sigprof I get the same error message when using both the 22.05 stable nixpkgs and *today's* version:

```nix
{ pkgs ? import
    (
      fetchTarball
        {
          name = "nixpkgs-22.05-2022-06-13";
[...]

PR #642, opened 2022-05-16
Use flit-core as build system for pyparsing after 3.0.8

pmiddend:
The project switched to flit. See the [Release Notes for pyparsing](https://github.com/pyparsing/pyparsing/releases).

PR #663, opened 2022-06-15
build-systems: optional build system objects with version information

trepetti:
jsonschema switched to the hatchling backend starting with [version 4.6.0](https://github.com/python-jsonschema/jsonschema/releases/tag/v4.6.0).

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -591,6 +591,9 @@
   "json-schema-for-humans": [
     "poetry-core"
   ],
+  "jsonschema": [
+    "hatchling"
+  ],
   "jupyter-server": [
     "jupyter-packaging"
   ],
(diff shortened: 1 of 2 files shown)


trepetti:
@adisbladis let me know if you like this extension to the schema of `build-systems.json`. It takes advantage of the mixed-type capability of JSON/Nix arrays/lists to optionally let you put a build system dependency in JSON as an object/attr set instead of just a string. This fixes the behavior seen in https://github.com/nix-community/poetry2nix/issues/655 by eliminating an internal inconsistency in package versions and their build systems. It also could be used to let https://github.com/nix-community/poetry2nix/pull/643 build correctly and achieve version-conditional build system specification in JSON like what was originally being proposed in https://github.com/nix-community/poetry2nix/pull/642 in `default.nix`.

adisbladis:
This looks great! Thank you very much!

PR #664, opened 2022-06-15
build-systems: add sane-python

considerate:
`sane-python` uses `poetry-core` as the build-system.


trepetti:
I ran into this issue as well. CI connecting to Cachix was being a bit flaky this afternoon.

PR #666, opened 2022-06-15
Use hatchling build system for jsonschema

l0b0:


l0b0:
Does anyone know how to deal with the error?

> Could not find a version that satisfies the requirement pluggy>=1.0.0; python_version > "3" (from hatchling) (from versions: none)

trepetti:
@l0b0 https://github.com/nix-community/poetry2nix/pull/663 would fix this (it originally started as a commit to add hatchling to jsonschema too). The issue is that while there will only be one true version of every package for the `buildInputs` that get imported into the Python interpreter at application runtime, the `nativeBuildInputs` like the build system used at build time, could potentially have conflicts with one another depending on the dependency versions each build system uses which can lead to irreconcilable conflicts like you are seeing here. Adding the ability to specify the versions at which a given build system should be use resolve the problem. https://github.com/nix-community/poetry2nix/pull/663 is able to get jsonschema to build fine.

PR #687, opened 2022-07-23
Don't apply the mypy patches after 0.971, the changes are released.

cyraxjoe:
[Changelog](https://mypy-lang.blogspot.com/2022/07/mypy-0971-released.html) with the included pull request https://github.com/python/mypy/pull/11143 in mypy.


cyraxjoe:
About the `FIXME` [comment](https://github.com/nix-community/poetry2nix/blob/f839d908f0b1e774e969aceccd7d7f0e451ef35f/overrides/default.nix#L1144), the patches are still useful if for some reason the user in using any of those version in between `0.900  < 0.971` but I'm not sure if the _policy_ is to only supported the latest, on which case I can just remove the patches and that comment.

Issue #284, opened 2021-04-12
Cannot reference package using {path = "../<package>", develop = true}


confususs:
First I want to say thank you for all the hard work on this project! I really love it & I'm trying to transfer all my poetry projects to it. :smile: 

Currently I'm using poetry within a nix-shell, where I have the following layout:

```
├── consumer
│   ├── consumer
│   │   └── main.py
│   └── pyproject.toml
├── core
│   ├── core
│   │   └── models.py
│   └── pyproject.toml
├── poetry.lock
├── pyproject.toml
└── shell.nix
```

You can check it out @ https://github.com/confususs/poetry2nix-example. 

The idea is to be able to have 1 'core package' which can be consumed by multiple packages. Each package has its own `pyproject.toml` & `poetry.lock`. 

On top of that, I've defined another `pyproject.toml` & `poetry.lock` on the top level of the repository so that I can easily access all my installed packages within my IDE once I called `poetry install` after calling `nix-shell`. This seems to work fine. (to reproduce this, uncomment `myAppEnv` in the shell.nix)

However, I'd like to use poetry2nix to automatically install all my dependencies for me once I enter my nix-shell, but the configuration is deemed incorrect somehow. See the output:

```
poetry2nix-example on  main is 📦 v0.1.0 via 🐍 v3.8.8 (python-3.8.8)took 4s
➜ nix-shell
these derivations will be built:
  /nix/store/mkw9lzys7xypd4xnqcr22dl23pr2m7b9-python3.8-consumer-0.1.0.drv
  /nix/store/2b6irdl3w72nbjvr7a383fy4mj2rv96s-python3-3.8.8-env.drv
building '/nix/store/mkw9lzys7xypd4xnqcr22dl23pr2m7b9-python3.8-consumer-0.1.0.drv'...
Sourcing python-remove-tests-dir-hook
Sourcing python-catch-conflicts-hook.sh
Sourcing python-remove-bin-bytecode-hook.sh
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
Sourcing pip-install-hook
Using pipInstallPhase
Sourcing python-imports-check-hook.sh
Using pythonImportsCheckPhase
Sourcing python-namespaces-hook
unpacking sources
unpacking source archive /nix/store/g7r0v3d6s3mircyisc67bsq3pgjllqyw-source
source root is source
setting SOURCE_DATE_EPOCH to timestamp 315619200 of file source/pyproject.toml
patching sources
configuring
[...]

PR #689, opened 2022-07-29
Fix build failure when pyproject.toml contains editable path dependencies

crossing:
`mkPoetryApplication` fails when a editable path dependency exists in `pyproject.toml`, similar to #284 

This PR fix the problem by removing `develop` attributes in path / git dependencies in hooks.

### DIFF
--- a/hooks/pyproject-without-special-deps.py
+++ b/hooks/pyproject-without-special-deps.py
@@ -22,6 +22,7 @@ def main(input, output, fields_to_remove):
                     any_removed |= dep.pop(field, None) is not None
                 if any_removed:
                     dep["version"] = "*"
+                    dep.pop("develop", None)
 
     output.write(tomlkit.dumps(data))
 
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1351,13 +1351,18 @@ lib.composeManyExtensions [
       );
 
       pillow = super.pillow.overridePythonAttrs (
-        old: {
+        old:
+        let
+          preConfigure = (old.preConfigure or "") + pkgs.python3.pkgs.pillow.preConfigure;
+        in
+        {
           nativeBuildInputs = (old.nativeBuildInputs or [ ])
             ++ [ pkg-config self.pytest-runner ];
           buildInputs = with pkgs; (old.buildInputs or [ ])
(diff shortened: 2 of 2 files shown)


crossing:
I don't have a mac to check the failed job. And the message doesn't seem to be related. Any hints what I should do?

sigprof:
The issues with the `pillow` package build on darwin probably appeared a long time before, but remain hidden in most cases, because the package does not get rebuilt. Probably your change touched something that caused a rebuild of that package and uncovered the breakage.

The underlying reason is that the build scripts of the `pillow` package try to detect the presence of various optional libraries, but do that in a way that is not really compatible with Nix; apparently at some point the GitHub `macos-latest` environment changed somehow, and those scripts started to detect the presence of the `jpeg2000` library in there (installed through homebrew, probably as a dependency of something else), even though that library is not in Nix store, therefore the build process fails later when it attempts to use the library outside of the Nix store.

I'm currently using this override for `pillow`, which allows it to build on GitHub CI for `macos-latest` (but I tested only the latest version, which happens to be 9.2.0 at the moment, while the CI job that failed here apparently tried to build 6.2.2 — no idea whether this override would actually work for that version):
```nix
      pillow = super.pillow.overridePythonAttrs(old: {
        # Use preConfigure from nixpkgs to fix library detection issues and
        # impurities which can break the build process; this also requires
        # adding propagatedBuildInputs and buildInputs from the same source.
        propagatedBuildInputs = (old.buildInputs or []) ++ pkgs.python3.pkgs.pillow.propagatedBuildInputs;
        buildInputs = (old.buildInputs or []) ++ pkgs.python3.pkgs.pillow.buildInputs;
        preConfigure = (old.preConfigure or "") + pkgs.python3.pkgs.pillow.preConfigure;
      });
```
But I'm not sure whether this is the right solution (maybe the correct thing would be to copy [the required parts from nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/pillow/generic.nix) instead of extracting them from the package like that, but they are rather extensive). I also don't have a Mac and can use only the GitHub CI for testing the Mac-specific changes.

crossing:
Thanks @sigprof 

Made a change to use `preConfigure` from nixpkg's pillow and now test passed.

Pushed an empty commit to trigger action re-runs.

Let me know if I should stash the commits.

PR #702, opened 2022-08-30
fix poetry replacing underscores with hyphens

vale981:
Poetry replaces underscores with hyphens in module names.

This leads to the problem that things like
```

[tool.poetry.dependencies]
hiro_models = { git = "git@github.com:vale981/two_qubit_model.git", branch="online_calculation" }
```
don't work, because the `sourceSpec` is empty in

```
if isGit then
          (
            builtins.fetchGit ({
              inherit (source) url;
              rev = source.resolved_reference or source.reference;
              ref = sourceSpec.branch or (if sourceSpec ? tag then "refs/tags/${sourceSpec.tag}" else "HEAD");
            } // (
              let
                nixVersion = builtins.substring 0 3 builtins.nixVersion;
              in
              lib.optionalAttrs ((sourceSpec ? rev) && (lib.versionAtLeast nixVersion "2.4")) {
                allRefs = true;
              }
            ))
          )
```

The poetry lock entry for this looks like
```
[[package]]
name = "hiro-models"
version = "1.2.3"
description = "Operators for a general model of two interacting qubits coupled to two baths."
category = "main"
optional = false
python-versions = ">=3.9,<3.11"
develop = false
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -7,7 +7,7 @@ let
   # Poetry2nix version
   version = "1.31.0";
 
-  inherit (poetryLib) isCompatible readTOML moduleName;
+  inherit (poetryLib) isCompatible readTOML moduleName underscorify;
 
   # Map SPDX identifiers to license names
   spdxLicenses = lib.listToAttrs (lib.filter (pair: pair.name != null) (builtins.map (v: { name = if lib.hasAttr "spdxId" v then v.spdxId else null; value = v; }) (lib.attrValues lib.licenses)));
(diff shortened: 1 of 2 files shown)


vale981:
Thanks for merging ;). I wasn't sure whether this is the correct way to do it, but apparently it is.

sciyoshi:
This change causes a regression: determining sourceSpec breaks if the dependency is specified in the pyproject.toml using dashes instead of underscores, which is allowed.

sciyoshi:
In fact, poetry prefers dashes to underscores, as you can see below:
```
~/workspace/oss/test is 📦 v0.1.0 via 🐍 v3.9.13
λ poetry add typing_extensions
Using version ^4.3.0 for typing-extensions

Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing typing-extensions (4.3.0)

~/workspace/oss/test is 📦 v0.1.0 via 🐍 v3.9.13
λ cat pyproject.toml
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["Samuel Cormier-Iijima <samuel@cormier-iijima.com>"]

[tool.poetry.dependencies]
python = "^3.10"
typing-extensions = "^4.3.0"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
```
This leads me to think that this change should be reverted until a proper fix can be put in place.

Issue #681, opened 2022-07-06
Poetry 1.2.0b2


lelit:
Is there a plan to update "vendored" poetry to 1.2.0b2?

I ask because I'm facing what seems a [resolver problem with poetry 1.1.13](https://github.com/python-poetry/poetry/discussions/5936), apparently already fixed in poetry 1.2.0b2, that is preventing upgrades in a poetry2nix environment I'm working.

While I was able to adapt the [nixpkgs expression](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/poetry/default.nix) to 1.0.2b2, basically overriding some of its dependencies like
```nix
  cleo' = pypkgs.buildPythonPackage rec {
    pname = "cleo";
    version = "1.0.0a5";
    src = pypkgs.fetchPypi {
      inherit pname version;
      sha256 = "sha256-CXydDgMy/VPMifwR6wproDCeajkzwI97OFWFVUhpJdM=";
    };
    propagatedBuildInputs = [pypkgs.clikit];
    doCheck = false;
  };

  poetry-core' = pypkgs.poetry-core.overridePythonAttrs (old: rec {
    version = "1.1.0b2";
    src = fetchFromGitHub {
      owner = "python-poetry";
      repo = "poetry-core";
      rev = version;
      sha256 = "sha256-7H5J0S7euh3tZ6Ug70WOgMm7yIxDBpVCFztpGqv5VMs=";
    };
    doCheck = false;
  });

  poetry-plugin-export' = pypkgs.buildPythonPackage rec {
    pname = "poetry-plugin-export";
    version = "1.0.4";
    src = pypkgs.fetchPypi {
      inherit pname version;
      hash = "sha256-38i5sScXz/JP1fdDRs811Kz5eYHXCr7oX2LK6kFUGhs=";
    };
    nativeBuildInputs = [pypkgs.pythonRelaxDepsHook];
    pythonRemoveDeps = ["poetry"];
    buildInputs = [
      poetry-core'
    ];
    doCheck = false;
  };
```
and little else, my Nix knowledge was not enough of update poetry2nix `pkgs/poetry`.

Thanks for any feedback.

lelit:
Wrt to the problem that originated the request, I [tracked down](https://github.com/python-poetry/poetry/discussions/5936#discussioncomment-3113135) the problem to `poetry-core` 1.0.x

takeda:
[...]

PR #704, opened 2022-09-01
poetry: 1.1.15 -> 1.2.0

adisbladis:
This contains support for the update Poetry 1.2.0 `pyproject.toml` schema while still retaining backwards compatibility with any older files.

Depends on https://github.com/NixOS/nixpkgs/pull/187999.
Once that + the backport are in released channels this PR will be merged.

----

Closes https://github.com/nix-community/poetry2nix/issues/681


yajo (approved):
Code looks good.

Smaug123:
@adisbladis Per https://nixpk.gs/pr-tracker.html?pr=187999 the nixpkgs PR is now in nixos-unstable; per https://nixpk.gs/pr-tracker.html?pr=189283 the backport is in nixos-22.05. How's it looking for this PR?

cpcloud:
Well, time to try this out again!

cpcloud:
It looks like `geopandas` and `ipdb` are two packages that cannot be installed with poetry2nix due to an infinite recursion error, which appears to be because they both depend on `setuptools` as `install_requires` (runtime) dependencies.

I've actually had two PRs open for a bit on both of these projects to fix the issue (https://github.com/gotcha/ipdb/pull/254 and https://github.com/Toblerity/Fiona/pull/1139).

We were using `ipdb` in ibis as a development dependency, but now we're not because of that issue and we have `geopandas` which depends on `Fiona` as an optional dependency for geospatial functionality.

What's interesting is that the Fiona issue doesn't show up with `poetry < 1.2`.

Is there any way we can work around this issue in `poetry2nix`? Happy to open a PR, but not entirely sure where to start poking around.

cpcloud:
Another package that fails is `poetry-dynamic-versioning` with the same infinite recursion error, but I believe a different cycle. In `ipdb` and `geopandas` the cycle is between setuptools-scm and setuptools, where as here I believe the cycle is between poetry-dynamic-versioning and poetry (though I'm not 100% sure about that).

PR #706, opened 2022-09-05
Override for openvino

tfmoraes:


adisbladis:
I fixed up your commit message and squashed. LGTM.

PR #710, opened 2022-09-09
PEP 425 compatibility improvements

### LINKED ISSUE #254: PEP425 wheel matching is too strict
We should match older python versions, not just the current one.
See the priority list in PEP425: https://www.python.org/dev/peps/pep-0425/#id1.

In particular this was a problem with `cryptography` after https://github.com/pyca/cryptography/issues/5771 where the source build didn't work and setting `preferWheel` didn't match any wheels, ending up falling back to a source build. 

The wheel file `cryptography-3.4.4-cp36-abi3-manylinux2014_x86_64.whl` should not be filtered out as this is guaranteed to work with newer versions of CPython (as indicated by `abi3`).

vlaci:
Version specifiers should be interpreted as lower bounds for required interpreter version.

A side-effect of this change is that `cryptography` can be installed from a wheel.

Fixes #254

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -362,8 +362,9 @@ lib.composeManyExtensions [
             lib.warn "Unknown cryptography version: '${version}'. Please update getCargoHash." lib.fakeHash
           );
           sha256 = getCargoHash super.cryptography.version;
+          isWheel = lib.hasSuffix ".whl" super.cryptography.src;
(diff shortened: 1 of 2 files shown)


vlaci:
The build failed because of a network error in GitHub CI.

adisbladis:
Thank you!

PR #713, opened 2022-09-16
Add idna flit-core override

farcaller:
via https://github.com/kjd/idna/blob/master/pyproject.toml#L3


takeda:
It looks like this PR https://github.com/nix-community/poetry2nix/pull/715 already included flit-core for idna.

PR #716, opened 2022-09-20
Add override for idna 3.4+

K900:


takeda:
It looks like this PR https://github.com/nix-community/poetry2nix/pull/715 already included flit-core for idna.

Issue #721, opened 2022-09-25
Source packages not getting included in `mkPoetryEnv`


Smaug123:
### Describe the issue

It appears that `git` source packages incorrectly fail to end up in the `mkPoetryEnv` environment.

### Additional context

I'm attempting to build Stable Diffusion, via my fork; flake at https://github.com/invoke-ai/InvokeAI/blob/d87faddfcf37b3d44a04ae08e05554e78cac4a2b/flake.nix

At runtime, Stable Diffusion fails to start, with `ModuleNotFoundError: No module named 'taming'`. `taming` comes from `taming-transformers`, which is [in my poetry.lock](https://github.com/invoke-ai/InvokeAI/blob/b00b80808344f92c553f0e01e2d8ef83efe81e4f/poetry.lock#L2243); it's [obtained by source](https://github.com/invoke-ai/InvokeAI/blob/b00b80808344f92c553f0e01e2d8ef83efe81e4f/pyproject.toml#L133).

The resulting Python environment's `ls -la /nix/store/4li87c09d8fzddzib68rhkx70wmkk94n-python3-3.9.13-env/lib/python3.9/site-packages` contains a dist-info for `taming-transformers` (specifically `taming_transformers-0.0.1.dist-info -> /nix/store/2q3s7rk41rxphfm8ily7a8mr0qrpvxv6-python3.9-taming-transformers-0.0.1/lib/python3.9/site-packages/taming_transformers-0.0.1.dist-info`), but it does *not* contain the source anywhere:

```
> find /nix/store/4li87c09d8fzddzib68rhkx70wmkk94n-python3-3.9.13-env -name '*taming*'
/nix/store/4li87c09d8fzddzib68rhkx70wmkk94n-python3-3.9.13-env/lib/python3.9/site-packages/taming_transformers-0.0.1.dist-info
```

By contrast, a manually-created virtual environment has the source, checked out into `venv/src/taming-transformers`.

Smaug123:
This was actually an upstream bug; it's fixed by https://github.com/CompVis/taming-transformers/pull/173, if they ever decide to merge it.

PR #723, opened 2022-09-25
Overrides: ftfy, shortuuid, contourpy, omegaconf, opencv-python-headless

Smaug123:
These are part of the dependency footprint of Stable Diffusion, although in fact in real life I ended up overriding `opencv-python-headless` to just point to `opencv-python` (because in the particular case of Stable Diffusion, both packages are in the footprint and they conflict).

I've tested these build on aarch64-darwin, although I haven't yet got the end-to-end test (namely "run SD") because of at least two unrelated problems (#721, and my inability to get the libjpeg dylib into the `torchvision` search path).

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -435,6 +435,9 @@
   "frozenlist": [
     "cython"
   ],
+  "ftfy": [
+    "poetry"
+  ],
   "funcparserlib": [
     "poetry-core"
   ],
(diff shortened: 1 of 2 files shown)


Smaug123:
The test cancellation was a timeout (https://github.com/nix-community/poetry2nix/actions/runs/3122436605/jobs/5064394180).

PR #729, opened 2022-09-27
Overrides: add suds, mdit-py-plugins

rszamszur:
Overrides added:
1) suds - For some reason, suds deploy the same package under two different names, where the name is configured at build time via environment variable. https://github.com/suds-community/suds/blob/a616d96b070ca119a532ff395d4a2a2ba42b257c/setup.py#L648
2) mdit-py-plugins - https://github.com/executablebooks/mdit-py-plugins/blob/c35bf143fb45e8572a1488cd78a96a1bbbb92bfb/pyproject.toml#L2

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -740,6 +740,9 @@
   "mdformat": [
     "poetry-core"
   ],
+  "mdit-py-plugins": [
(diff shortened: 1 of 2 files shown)


adisbladis:
thank you very much!

Issue #616, opened 2022-05-08
mkPoetryEnv does not create egg-info metadata correctly


schlarpc:
### Describe the issue

When executing a package built with `mkPoetryApplication`, `importlib.metadata` is able to access the built package's metadata such as the version. When operating under `mkPoetryEnv`, the metadata is not found by Python. This appears to be caused by misnaming of `.egg-info` metadata files.

#### Relevant files

- `shell.nix`
```
{ pkgs ? import <nixpkgs> {} }:
let
  myAppEnv = pkgs.poetry2nix.mkPoetryEnv {
    projectDir = ./.;
    editablePackageSources = {
      p2n-egginfo = ./src;
    };
  };
in myAppEnv.env
```

- `default.nix`
```
{ pkgs ? import <nixpkgs> {} }:
let
  myApp = pkgs.poetry2nix.mkPoetryApplication {
    projectDir = ./.;
  };
in myApp
```

- `pyproject.toml`
```
[tool.poetry]
name = "p2n-egginfo"
version = "0.1.0"
description = ""
authors = ["Chaz Schlarp <schlarpc@gmail.com>"]

[tool.poetry.dependencies]
python = "^3.9"

[tool.poetry.dev-dependencies]

[tool.poetry.scripts]
p2n-egginfo-test = "p2n_egginfo:main"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

- `poetry.lock`
```
package = []

[metadata]
lock-version = "1.1"
python-versions = "^3.9"
content-hash = "ce2aa767160f871dd3652615ba0a0dceb7733d62eb8cb4665b87f30a562e3adf"

[metadata.files]
```

- `src/p2n_egginfo/__init__.py`
```
import importlib.metadata

def main():
    print(importlib.metadata.version('p2n_egginfo'))
```

#### Explanation

Under `mkPythonApplication`, accessing the package metadata works fine:

```
$ nix-build; and ./result/bin/p2n-egginfo-test
/nix/store/qr6ijk5ck6d1rsi73fci9jdd1s1jjc0m-python3.9-p2n-egginfo-0.1.0
0.1.0
```

But under `mkPoetryEnv`, it fails. 

```
$ nix-shell --run 'p2n-egginfo-test'
Traceback (most recent call last):
[...]

PR #732, opened 2022-09-29
Generate a dist-info folder instead of an egg-info

silky:
All we really need is a `METADATA` file, and for the folder to have dashes replaced with underscores.

This may be related to https://github.com/nix-community/poetry2nix/issues/616.

I think this probably shouldn't be merged as-is, because I'm just throwing away the egg part, but @adisbladis if you are pro implementing this at all we could:

1. Introduce a new `editablePackagesSources` argument to decide whether to use egg or dist-info style,
2. Figure out what information is essential, and add that too,
3. ???

Alternatively, this PR by itself might help people that are looking for `dist-info` files, which get picked up by libraries such as `pkg_resources.get_distribution("your-package")`.

### DIFF
--- a/editable.nix
+++ b/editable.nix
@@ -7,6 +7,8 @@
 }:
 let
   name = poetryLib.normalizePackageName pyProject.tool.poetry.name;
+  underscoredName = builtins.replaceStrings [ "-" ] [ "_" ] name;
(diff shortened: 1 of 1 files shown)


silky:
any thoughts on what would would be required to get this merged @K900 ? would be nice to not have to keep the fork up-to-date :)

PR #736, opened 2022-09-30
Trying to fix the setuptools fallout

K900:
TODO list (ongoing):

* [x] fix package name normalization
* [x] fix override updating script to not screw with existing overrides with version constraints

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -7,7 +7,7 @@ let
   # Poetry2nix version
   version = "1.31.0";
 
-  inherit (poetryLib) isCompatible readTOML moduleName underscorify;
+  inherit (poetryLib) isCompatible readTOML normalizePackageName normalizePackageSet;
 
   # Map SPDX identifiers to license names
   spdxLicenses = lib.listToAttrs (lib.filter (pair: pair.name != null) (builtins.map (v: { name = if lib.hasAttr "spdxId" v then v.spdxId else null; value = v; }) (lib.attrValues lib.licenses)));
@@ -16,6 +16,10 @@ let
 
   # Experimental withPlugins functionality
   toPluginAble = (import ./plugins.nix { inherit pkgs lib; }).toPluginAble;
+
+  # List of known build systems that are passed through from nixpkgs unmodified
+  knownBuildSystems = builtins.fromJSON (builtins.readFile ./known-build-systems.json);
+  nixpkgsBuildSystems = lib.subtractLists [ "poetry" "poetry-core" ] knownBuildSystems;
 
   mkInputAttrs =
     { py
@@ -38,7 +42,7 @@ let
             (
               dep:
               let
-                pkg = py.pkgs."${moduleName dep}";
+                pkg = py.pkgs."${normalizePackageName dep}";
                 constraints = depSet.${dep}.python or "";
                 isCompat = compat constraints;
               in
@@ -59,7 +63,6 @@ let
       buildInputs = mkInput "buildInputs" (if includeBuildSystem then buildSystemPkgs else [ ]);
       propagatedBuildInputs = mkInput "propagatedBuildInputs" (
         (getDeps pyProject.tool.poetry."dependencies" or { })
-        ++ ([ py.pkgs.setuptools ])
         ++ (
           # >=poetry-1.2.0 dependency groups
           if pyProject.tool.poetry.group or { } != { }
@@ -128,7 +131,7 @@ lib.makeScope pkgs.newScope (self: {
     , editablePackageSources ? { }
     , pyProject ? readTOML pyproject
     , groups ? [ ]
-    }@attrs:
+    }:
     let
       /* The default list of poetry2nix override overlays */
       mkEvalPep508 = import ./pep508.nix {
@@ -157,15 +160,7 @@ lib.makeScope pkgs.newScope (self: {
         let
           lockfiles = lib.getAttrFromPath [ "metadata" "files" ] poetryLock;
         in
-        lib.listToAttrs (lib.mapAttrsToList (n: v: { name = moduleName n; value = v; }) lockfiles);
-      specialAttrs = [
-        "overrides"
-        "poetrylock"
-        "projectDir"
-        "pwd"
-        "preferWheels"
-      ];
-      passedAttrs = builtins.removeAttrs attrs specialAttrs;
+        lib.listToAttrs (lib.mapAttrsToList (n: v: { name = normalizePackageName n; value = v; }) lockfiles);
       evalPep508 = mkEvalPep508 python;
 
       # Filter packages by their PEP508 markers & pyproject interpreter version
@@ -183,12 +178,15 @@ lib.makeScope pkgs.newScope (self: {
       # closure as python can only ever have one version of a dependency
       baseOverlay = self: super:
         let
-          getDep = depName: self.${depName};
           lockPkgs = builtins.listToAttrs (
             builtins.map
               (
-                pkgMeta: rec {
-                  name = moduleName pkgMeta.name;
+                pkgMeta:
+                if builtins.elem pkgMeta.name nixpkgsBuildSystems then {
+                  name = pkgMeta.name;
+                  value = super."${pkgMeta.name}";
+                } else rec {
+                  name = normalizePackageName pkgMeta.name;
                   value = self.mkPoetryDep (
                     pkgMeta // {
                       inherit pwd preferWheels;
@@ -196,17 +194,16 @@ lib.makeScope pkgs.newScope (self: {
                       files = lockFiles.${name};
                       pythonPackages = self;
 
-                      # Packages can be specified with underscores in pyproject.toml; check for
-                      # both possibilities.
-                      sourceSpec = with pyProject.tool.poetry; (
-                        dependencies.${pkgMeta.name} or
-                          dependencies.${underscorify pkgMeta.name} or
-                            dev-dependencies.${pkgMeta.name} or
-                              dev-dependencies.${underscorify pkgMeta.name} or
-                                group.dev.dependencies.${underscorify pkgMeta.name} or # Poetry 1.2.0+
-                                  { }
-
-                      );
+                      sourceSpec =
+                        let
+                          normalizedName = normalizePackageName pkgMeta.name;
+                        in
+                        with pyProject.tool.poetry; (
+                          (normalizePackageSet dependencies).${normalizedName}
+                            or (normalizePackageSet dev-dependencies).${normalizedName}
+                            or (normalizePackageSet group.dev.dependencies).${normalizedName} # Poetry 1.2.0+
+                            or { }
+                        );
                     }
                   );
                 }
@@ -255,7 +252,7 @@ lib.makeScope pkgs.newScope (self: {
               super)
 
             # Null out any filtered packages, we don't want python.pkgs from nixpkgs
-            (self: super: builtins.listToAttrs (builtins.map (x: { name = moduleName x.name; value = null; }) incompatible))
+            (self: super: builtins.listToAttrs (builtins.map (x: { name = normalizePackageName x.name; value = null; }) incompatible))
             # Create poetry2nix layer
             baseOverlay
 
@@ -303,7 +300,7 @@ lib.makeScope pkgs.newScope (self: {
     , groups ? [ "dev" ]
     }:
     let
-      inherit (lib) elem hasAttr;
+      inherit (lib) hasAttr;
 
       pyProject = readTOML pyproject;
 
@@ -395,7 +392,7 @@ lib.makeScope pkgs.newScope (self: {
             py.pkgs.removeGitDependenciesHook
           ];
         } // {
-          pname = moduleName pyProject.tool.poetry.name;
+          pname = normalizePackageName pyProject.tool.poetry.name;
           version = pyProject.tool.poetry.version;
 
           inherit src;
@@ -477,7 +474,7 @@ lib.makeScope pkgs.newScope (self: {
 
     Can be overriden by calling defaultPoetryOverrides.overrideOverlay which takes an overlay function
   */
-  defaultPoetryOverrides = self.mkDefaultPoetryOverrides (import ./overrides { inherit pkgs lib; });
+  defaultPoetryOverrides = self.mkDefaultPoetryOverrides (import ./overrides { inherit pkgs lib poetryLib; });
 
   /*
     Convenience functions for specifying overlays with or without the poerty2nix default overrides
--- a/editable.nix
+++ b/editable.nix
@@ -6,7 +6,7 @@
 , editablePackageSources
 }:
 let
-  name = poetryLib.moduleName pyProject.tool.poetry.name;
+  name = poetryLib.normalizePackageName pyProject.tool.poetry.name;
 
   # Just enough standard PKG-INFO fields for an editable installation
   pkgInfoFields = {
--- a/known-build-systems.json
+++ b/known-build-systems.json
@@ -1,0 +1,12 @@
+[
+    "poetry",
+    "poetry-core",
+    "flit",
+    "flit-core",
+    "pbr",
+    "flitBuildHook",
+    "cython",
+    "hatchling",
+    "setuptools",
+    "setuptools-scm"
+]
--- a/lib.nix
+++ b/lib.nix
@@ -8,12 +8,16 @@ let
     genList (i: if i == idx then value else (builtins.elemAt list i)) (length list)
   );
 
-  # Do some canonicalisation of module names
-  moduleName = name: lib.toLower (lib.replaceStrings [ "_" "." ] [ "-" "-" ] name);
+  # Normalize package names as per PEP 503
+  normalizePackageName = name:
+    let
+      parts = builtins.split "[-_.]+" name;
+      partsWithoutSeparator = builtins.filter (x: builtins.typeOf x == "string") parts;
+    in
+    lib.strings.toLower (lib.strings.concatStringsSep "-" partsWithoutSeparator);
 
-  # For some reason, poetry replaces underscores with dashes in module
-  # names, this has to be reversed sometimes
-  underscorify = name: (lib.replaceStrings [ "-" ] [ "_" ] name);
+  # Normalize an entire attrset of packages
+  normalizePackageSet = lib.attrsets.mapAttrs' (name: value: lib.attrsets.nameValuePair (normalizePackageName name) value);
 
   # Get a full semver pythonVersion from a python derivation
   getPythonVersion = python:
@@ -237,8 +241,8 @@ in
     getBuildSystemPkgs
     satisfiesSemver
     cleanPythonSources
-    moduleName
-    underscorify
+    normalizePackageName
+    normalizePackageSet
     getPythonVersion
     getTargetMachine
     ;
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -26,7 +26,7 @@ pythonPackages.callPackage
     }@args:
     let
       inherit (python) stdenv;
-      inherit (poetryLib) isCompatible getManyLinuxDeps fetchFromLegacy fetchFromPypi moduleName;
+      inherit (poetryLib) isCompatible getManyLinuxDeps fetchFromLegacy fetchFromPypi normalizePackageName;
 
       inherit (import ./pep425.nix {
         inherit lib poetryLib python stdenv;
@@ -88,23 +88,10 @@ pythonPackages.callPackage
             else (builtins.elemAt (lib.strings.splitString "-" name) 2);
         };
 
-      # Prevent infinite recursion
-      skipSetupToolsSCM = [
-        "setuptools_scm"
-        "setuptools-scm"
-        "toml" # Toml is an extra for setuptools-scm
-        "tomli" # tomli is an extra for later versions of setuptools-scm
-        "flit-core"
-        "packaging"
-        "six"
-        "pyparsing"
-        "typing-extensions"
-      ];
-      baseBuildInputs = lib.optional (! lib.elem name skipSetupToolsSCM) pythonPackages.setuptools-scm;
       format = if isDirectory || isGit || isUrl then "pyproject" else fileInfo.format;
     in
     buildPythonPackage {
-      pname = moduleName name;
+      pname = normalizePackageName name;
       version = version;
 
       inherit format;
@@ -124,10 +111,9 @@ pythonPackages.callPackage
       ];
 
       buildInputs = (
-        baseBuildInputs
-        ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) pythonPackages.setuptools
-        ++ lib.optional (!isSource) (getManyLinuxDeps fileInfo.name).pkg
+        lib.optional (!isSource) (getManyLinuxDeps fileInfo.name).pkg
         ++ lib.optional isDirectory buildSystemPkgs
+        ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) pythonPackages.setuptools
       );
 
       propagatedBuildInputs =
@@ -149,7 +135,7 @@ pythonPackages.callPackage
             );
           depAttrs = lib.attrNames deps;
         in
-        builtins.map (n: pythonPackages.${moduleName n}) depAttrs;
+        builtins.map (n: pythonPackages.${normalizePackageName n}) depAttrs;
 
       meta = {
         broken = ! isCompatible (poetryLib.getPythonVersion python) python-versions;
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -1,1229 +1,12131 @@
 {
-  "HTSeq": [
-    "cython"
+  "aadict": [
+    "setuptools"
   ],
-  "PyStemmer": [
-    "cython"
+  "aafigure": [
+    "setuptools"
+  ],
+  "aardwolf": [
+    "setuptools"
+  ],
+  "abodepy": [
+    "setuptools"
+  ],
+  "absl-py": [
+    "setuptools"
+  ],
+  "absolufy-imports": [
+    "setuptools"
+  ],
+  "accuweather": [
+    "setuptools"
+  ],
+  "acme": [
+    "setuptools"
+  ],
+  "acme-tiny": [
+    "setuptools",
+    "setuptools-scm"
+  ],
+  "acoustics": [
+    "setuptools"
+  ],
+  "actdiag": [
+    "setuptools"
+  ],
+  "adafruit-io": [
+    "setuptools",
+    "setuptools-scm"
+  ],
+  "adafruit-nrfutil": [
+    "setuptools"
+  ],
+  "adafruit-platformdetect": [
+    "setuptools",
+    "setuptools-scm"
+  ],
+  "adafruit-pureio": [
+    "setuptools",
+    "setuptools-scm"
+  ],
+  "adal": [
+    "setuptools"
+  ],
+  "adax": [
+    "setuptools"
+  ],
+  "adax-local": [
+    "setuptools"
+  ],
+  "adb-enhanced": [
+    "setuptools"
+  ],
+  "adb-homeassistant": [
+    "setuptools"
+  ],
+  "adb-shell": [
+    "setuptools"
+  ],
+  "adblock": [
+    "setuptools"
+  ],
+  "add-trailing-comma": [
+    "setuptools"
+  ],
+  "addict": [
+    "setuptools"
+  ],
+  "adext": [
+    "setuptools",
+    "setuptools-scm"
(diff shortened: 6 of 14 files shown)


cpcloud:
@K900 Thanks for looking into this. I pushed a commit to the PR to format the code to get more CI to run.

cpcloud:
It looks like the pluggy dependency for hatchling isn't found because the hatchling version requires pluggy 1.0.0 but 0.13.1 is found.

cpcloud:
I'm trying this against a project I work on with complex dependencies (https://github.com/ibis-project/ibis) and this PR appears to still infinitely recurse, but in a way that isn't detectable until after evaluation has started (and eventually fails with a stack overflow), as opposed to `poetry2nix/master` in which the recursion is detected during evaluation and a somewhat useful error message is reported.

K900:
There's definitely new circular dependencies in there somewhere, we'll have to minimize it to find out.

cpcloud:
Using `poetry<1.2`, `setuptools` is never included in a package's dependencies in `poetry.lock`, whereas with `poetry==1.2` now packages that have a (likely incorrect) runtime dependency on setuptools have that dependency listed in the lock file.

Removing those dependencies' dependency on setuptools in the lockfile *and* removing `setuptools` itself from the lockfile allows evaluation to proceed.

cpcloud:
I have upstream'd a few PRs to help address the incorrect setuptools dependency in the ecosystem:

one to Toblerity/Fiona likely to be merged: https://github.com/Toblerity/Fiona/pull/1139
one to ipdb unlikely to be merged: https://github.com/gotcha/ipdb/pull/254
one to numba, not merged but addressed in another open PR: https://github.com/numba/numba/pull/8366

K900:
I'm still not sure why we're looping on setuptools at all tbh, especially now that there's no implicit dependency on it.

cpcloud:
I've pushed a commit to bring back the `from`/`until` version bounds on the build system overrides, since that was causing failures due to bringing in incompatible versions of transitive dependencies.

cpcloud:
I haven't yet been able isolate a smaller reproducible example than what's in ibis-project/ibis, but it appears there may be a loop in `python-remove-tests-dir-hook`?

When I run

```
nix build -f . -vvv --dry-run 2>&1 | tee /tmp/output.txt
```

on my `cpcloud/use-our-nixpkgs` branch of `ibis-project/ibis`

I see this output repeated ad infinitum:

```
instantiated 'python-remove-tests-dir-hook' -> '/nix/store/i1x50xj78xyv9zra1qpjb5dp42jqd7aj-python-remove-tests-dir-hook.drv'
instantiated 'python-setup-hook.sh' -> '/nix/store/bcjff6dn4gp7pf0knnja9ax4qrll78d5-python-setup-hook.sh.drv'
instantiated 'Python-3.10.7.tar.xz' -> '/nix/store/ap22y5ilipad2bhs555rldnir8hki1j2-Python-3.10.7.tar.xz.drv'
instantiated 'python3-3.10.7' -> '/nix/store/c5rlaqdfymd3j5v41rh98pjcv3dx3sh7-python3-3.10.7.drv'
instantiated 'hook' -> '/nix/store/z7br4z8im93lckrpxgwbrscfhwyh0ic8-hook.drv'
instantiated 'python-remove-tests-dir-hook' -> '/nix/store/i1x50xj78xyv9zra1qpjb5dp42jqd7aj-python-remove-tests-dir-hook.drv'
instantiated 'python-setup-hook.sh' -> '/nix/store/bcjff6dn4gp7pf0knnja9ax4qrll78d5-python-setup-hook.sh.drv'
instantiated 'Python-3.10.7.tar.xz' -> '/nix/store/ap22y5ilipad2bhs555rldnir8hki1j2-Python-3.10.7.tar.xz.drv'
instantiated 'python3-3.10.7' -> '/nix/store/c5rlaqdfymd3j5v41rh98pjcv3dx3sh7-python3-3.10.7.drv'
instantiated 'hook' -> '/nix/store/z7br4z8im93lckrpxgwbrscfhwyh0ic8-hook.drv'
```

cpcloud:
@K900 Hope you don't mind, but I pushed what I think is a minimal test case project: it only depends on `setuptools`. This is sufficient to cause the infinite recursion at eval time.

K900:
Please continue, I probably won't have much more time to look into this in the next few days.

default.nix
```diff
@@ -187,7 +187,11 @@ lib.makeScope pkgs.newScope (self: {
           lockPkgs = builtins.listToAttrs (
             builtins.map
               (
-                pkgMeta: rec {
+                pkgMeta:
+                if pkgMeta.name == "setuptools" then {
+                  name = pkgMeta.name;
+                  value = super.setuptools;
```
K900, line 193: This should probably be extended to cover all the build systems we support...

cpcloud:
I was able to get the test case I added to pass by always using `super.setuptools` if it's included as a dependency.

I'll admit I'm not entirely sure why this works.

My hunch is that there are two different versions of setuptools in the closure: one coming from `super`, and one coming from the poetry lock file.

@adisbladis Thoughts on this?

default.nix
cpcloud: Yeah, that makes sense. I'd like to see if I can get to the bottom of the reason why the infinite recursion is happening before doing that though.

One downside to this approach is that the version of whatever the build system is always pinned to whatever version/commit nixpkgs is pinned to. For setuptools that probably doesn't matter much, but for the newer build systems I'm not so sure.

K900:
I don't think that's avoidable, and it might be better than getting whatever random thing we end up with in the lockfile.

roberth:
I've blocked a `nixops_unstable` update because of this issue, but it's not clear to me what's happened around `setuptools`. Is it because of a poetry version update? (A PR description or issue link would have clarified this)

How do you think this situation should/will affect the poetry2nix-based packages in the upcoming NixOS release?

K900:
Nixpkgs' Python hooks used to propagate setuptools, but they don't anymore, so all the things that implicitly expected it to be there exploded.

cpcloud:
There's a second issue here too: poetry < 1.2 automatically removed setup tools from dependencies, whereas 1.2 doesn't resulting in infinite recursion.

PedroRegisPOAR:
### Maybe I have found an reproducer

#### 1. From poetry2nix master:
```bash
nix run --no-write-lock-file github:nix-community/poetry2nix#poetry -- --version
```

The locked URL:
```bash
# To get the locked URL run:
# nix flake metadata --no-write-lock-file github:nix-community/poetry2nix
nix run --no-write-lock-file github:nix-community/poetry2nix/c2ee667d2fd19f70a9ac256b81f2cfc248c4f04d#poetry -- --version
```

Results in:
```bash
warning: not writing modified lock file of flake 'github:nix-community/poetry2nix':
• Added input 'flake-utils':
    'github:numtide/flake-utils/c0e246b9b83f637f4681389ecabcb2681b4f3af0' (2022-08-07)
• Added input 'nixpkgs':
    'github:NixOS/nixpkgs/2380304ae4975ef0d6ba13b64f990e94e8c5fdbb' (2022-10-04)
error: builder for '/nix/store/avk4c6r2y4mixriryam1rgrrbkzfivmi-python3.10-packaging-21.3.drv' failed with exit code 2;
       last 10 log lines:
       >   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
       >   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
       >   File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
       >   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
       >   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
       >   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
       >   File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
       > ModuleNotFoundError: No module named 'setuptools'
       > 
       > 
       For full logs, run 'nix log /nix/store/avk4c6r2y4mixriryam1rgrrbkzfivmi-python3.10-packaging-21.3.drv'.
error: 1 dependencies of derivation '/nix/store/wb17wpkm1p8w00cajp960w8yinwh75xy-python3.10-poetry-1.2.0.drv' failed to build
```


#### 2. From K900/poetry2nix/hack

```bash
nix run --no-write-lock-file github:K900/poetry2nix/hack#poetry -- --version
```

Results in:
```bash
warning: not writing modified lock file of flake 'github:K900/poetry2nix/hack':
• Added input 'flake-utils':
    'github:numtide/flake-utils/c0e246b9b83f637f4681389ecabcb2681b4f3af0' (2022-08-07)
• Added input 'nixpkgs':
    'github:NixOS/nixpkgs/2380304ae4975ef0d6ba13b64f990e94e8c5fdbb' (2022-10-04)
Poetry (version 1.2.0)
```

So, it works!


The locked URL:
```bash
# To get the locked URL run:
# nix flake metadata --no-write-lock-file github:K900/poetry2nix/hack
nix run --no-write-lock-file github:K900/poetry2nix/fc3b734766d49695b34259d68e823f3baa732c07#poetry -- --version
```

### Extra metadata


```bash
nix run nixpkgs#nix-info -- --markdown
```

 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.0-48-generic, Ubuntu, 22.04.1 LTS (Jammy Jellyfish), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.10.2`
 - channels(vagrant): `"nixpkgs"`
 - nixpkgs: `/home/vagrant/.nix-defexpr/channels/nixpkgs`


About the machine that it was tested (an Ubuntu Vagrant virtual machine):
```bash
nix run nixpkgs#neofetch -- --json
```

Outputs:
```bash
{
    "OS": "Ubuntu 22.04.1 LTS x86_64",
    "Host": "KVM/QEMU (Standard PC (i440FX + PIIX, 1996) pc-i440fx-7.0)",
    "Kernel": "5.15.0-48-generic",
    "Uptime": "7 mins",
    "Packages": "626 (dpkg), 44 (nix-user), 3 (snap)",
    "Shell": "bash 5.1.16",
    "Resolution": "1024x768",
    "Terminal": "/dev/pts/0",
    "CPU": "Intel (Haswell, no TSX, IBRS) (4) @ 3.392GHz",
    "GPU": "Cirrus Logic QEMU Virtual Machine",
    "Memory": "331MiB / 968MiB",
    "Version": "7.1.0"
}
```

K900:
Yes, that reproduces the missing setuptools issue, but not the infinite recursion issue.

cpcloud:
I guess the question is whether the solution to fix the infinite recursion is the correct one. I'm happy to add the other build systems to the check if people think the approach is the way forward.

K900:
I'm testing something like that here, will push in a bit.

K900:
Aaand pushed, let's see how much it explodes now.

cpcloud:
@K900 Sweet! I'll try it out on ibis too.

K900:
And I just ran into another fun issue: some overlays are specified with dashes, and some are specified with underscores, and pip doesn't actually care, so we probably need to normalize _everything_.

cpcloud:
Hm, that should be happening as part of this block of code in `default.nix`:

````
                      sourceSpec = with pyProject.tool.poetry; (
                        dependencies.${pkgMeta.name} or
                          dependencies.${underscorify pkgMeta.name} or
                            dev-dependencies.${pkgMeta.name} or
                              dev-dependencies.${underscorify pkgMeta.name} or
                                group.dev.dependencies.${underscorify pkgMeta.name} or # Poetry 1.2.0+
                                  { }

                      );
```

cpcloud:
Oh, nevermind, that is not normalizing any output, just the input pkg name

K900:
I don't think it's enough, I just found a project that depends on `jaraco-classes` and ignores the override for `jaraco_classes`. Also, we should probably normalize towards dashes, as that's what pip actually uses.

cpcloud:
I also am finding a few more missing build system overrides. Mind if I push?

K900:
Go ahead.

cpcloud:
Don't wait on me, I'm finding a bunch by running with `--keep-going`. I'll push when I catch 'em all

K900:
OK, I finally got all the tests passing again. Still need to fix package name normalization though.

K900:
Mostly making a note for myself, the normalization rules are documented here: https://peps.python.org/pep-0503/#normalized-names

And should probably be applied globally.

K900:
So I just realized there's another problem. Even if we normalize all the package names in poetry2nix, we'll still have to somehow deal with nixpkgs attributes not being normalized, and given we want this to work with stable branches, we probably have to do it on the poetry2nix side...

cpcloud:
I am continuing to find a bunch of packages that need `setuptools`.

I think we should merge this soon and I will follow up with a PR to add the missing build systems. It's easy enough for me to workaround the missing build systems downstream.

K900:
I still need to fix pyproject.toml parsing to use normalized package names too...

roberth:
> So I just realized there's another problem. Even if we normalize all the package names in poetry2nix, we'll still have to somehow deal with nixpkgs attributes not being normalized, and given we want this to work with stable branches, we probably have to do it on the poetry2nix side...

What's stopping you from applying the normalization function to the Nixpkgs attrs?

Something like

```nix
let
  myNixpkgsPythonPackages =
    lib.mapAttrs'
      (name: value: { name = underscorify name; inherit value; })
      pkgs.pythonXYPackages;
in
<lots-of-defs>
```

K900:
That's what I want to do, yes, but I _also_ want to fix it in nixpkgs proper.

cpcloud:
Should we track the normalization upstream and fix that concurrently so that this patch can be merged without having to wait for a nixpkgs bump?

FRidh:
A quick work-around would be to still always include `setuptools`, basically having the situation as it was before the change in Nixpkgs.

K900:
That won't resolve the poetry 1.2.0 dependency cycle issue.

SemMulder:
> That won't resolve the poetry 1.2.0 dependency cycle issue.

I opened an ugly fix over at #744, before stumbling on this PR.

I'm not sure if the changes in https://github.com/NixOS/nixpkgs/pull/185815 also break that fix.

But I think that the proper fix for solving that dependency issue is something along the lines of (ripped from #744):
>A better approach would be to build the versions of setuptools and setuptools-scm specified in poetry.lock using whatever setuptools and setuptools-scm is in nixpkgs, and use setuptools and setuptools-scm from poetry.lock afterwards. However, I'm not sure how to best go about this, since it would require a separate Python environment.

@K900 any ideas on that ^? And that could/should be a separate PR from this one, no?

K900:
That's kind of irrelevant, because the versions in poetry.lock are runtime deps, not build time. We don't actually know what versions are required at build time, so the best we can do is assume the latest works and rely on overrides.

K900:
Yes and no. We have some concept of build dependencies in Nix, and the Python hooks handle it correctly, the main problem is that Poetry itself doesn't make the distinction so we can't infer which is which from the lockfile.

cpcloud:
Other than the merge conflicts :)

K900:
I still want to fix pyproject.toml normalization, but we can probably merge as is and fix later. I'll squash and rebase in a few hours.

K900:
OK, I've just squashed the whole thing down to one commit, because I have no idea how to break it down anymore. This should be good now.

cpcloud:
@adisbladis You want to review this? Otherwise I am inclined to merge it.

adisbladis:
thank you very much @K900 for dealing with this.

asymmetric:
Since this got merged (I confirmed by pinning the flake to this commit), I get

```
 nix build 
warning: Git tree '/home/asymmetric/code/foo/bar' is dirty
error: infinite recursion encountered

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:312:41:

          311|   */
          312|   nameValuePair = name: value: { inherit name value; };
             |                                         ^
          313|
```

<details><summary>trace</summary>

```
warning: Git tree '/home/asymmetric/code/foo/bar' is dirty
error: infinite recursion encountered

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:312:41:

          311|   */
          312|   nameValuePair = name: value: { inherit name value; };
             |                                         ^
          313|

       … while evaluating the attribute 'babel'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:312:41:

          311|   */
          312|   nameValuePair = name: value: { inherit name value; };
             |                                         ^
          313|

       … while evaluating 'addBuildSystem'

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:9:5:

            8|   addBuildSystem =
            9|     { self
             |     ^
           10|     , drv

       … from call site

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:81:21:

           80|       (attr: systems: builtins.foldl'
           81|         (drv: attr: addBuildSystem {
             |                     ^
           82|           inherit drv self attr;

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:81:15:

           80|       (attr: systems: builtins.foldl'
           81|         (drv: attr: addBuildSystem {
             |               ^
           82|           inherit drv self attr;

       … from call site

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:80:23:

           79|     lib.mapAttrs
           80|       (attr: systems: builtins.foldl'
             |                       ^
           81|         (drv: attr: addBuildSystem {

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:80:14:

           79|     lib.mapAttrs
           80|       (attr: systems: builtins.foldl'
             |              ^
           81|         (drv: attr: addBuildSystem {

       … from call site

       … while evaluating 'checkInPkgs'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:22:20:

           21|   # python-packages.nix.
           22|   checkInPkgs = n: alias: if builtins.hasAttr n super
             |                    ^
           23|                           then throw "Alias ${n} is still in python-packages.nix"

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:29:32:

           28|                              (removeRecurseForDerivations
           29|                               (checkInPkgs n alias)))
             |                                ^
           30|                      aliases;

       … while evaluating 'removeRecurseForDerivations'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:8:33:

            7|   # set to appear while listing all the packages available.
            8|   removeRecurseForDerivations = alias: with lib;
             |                                 ^
            9|       if alias.recurseForDerivations or false then

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:28:31:

           27|     lib.mapAttrs (n: alias: removeDistribute
           28|                              (removeRecurseForDerivations
             |                               ^
           29|                               (checkInPkgs n alias)))

       … while evaluating 'isDerivation'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:427:18:

          426|   */
          427|   isDerivation = x: x.type or null == "derivation";
             |                  ^
          428|

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:16:8:

           15|   removeDistribute = alias: with lib;
           16|     if isDerivation alias then
             |        ^
           17|       dontDistribute alias

       … while evaluating 'removeDistribute'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:15:22:

           14|   # sets from building on Hydra.
           15|   removeDistribute = alias: with lib;
             |                      ^
           16|     if isDerivation alias then

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:27:29:

           26|   mapAliases = aliases:
           27|     lib.mapAttrs (n: alias: removeDistribute
             |                             ^
           28|                              (removeRecurseForDerivations

       … while evaluating anonymous lambda

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/top-level/python-aliases.nix:27:22:

           26|   mapAliases = aliases:
           27|     lib.mapAttrs (n: alias: removeDistribute
             |                      ^
           28|                              (removeRecurseForDerivations

       … from call site

       … while evaluating 'isDerivation'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:427:18:

          426|   */
          427|   isDerivation = x: x.type or null == "derivation";
             |                  ^
          428|

       … from call site

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/default.nix:248:20:

          247|               (name: value: (
          248|                 if lib.isDerivation value && lib.hasAttr "overridePythonAttrs" value
             |                    ^
          249|                 then value.overridePythonAttrs (_: { doCheck = false; })

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/default.nix:247:22:

          246|             (self: super: lib.mapAttrs
          247|               (name: value: (
             |                      ^
          248|                 if lib.isDerivation value && lib.hasAttr "overridePythonAttrs" value

       … from call site

       … while evaluating the attribute 'babel'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:312:41:

          311|   */
          312|   nameValuePair = name: value: { inherit name value; };
             |                                         ^
          313|

       … while evaluating 'addBuildSystem'

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:9:5:

            8|   addBuildSystem =
            9|     { self
             |     ^
           10|     , drv

       … from call site

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:81:21:

           80|       (attr: systems: builtins.foldl'
           81|         (drv: attr: addBuildSystem {
             |                     ^
           82|           inherit drv self attr;

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:81:15:

           80|       (attr: systems: builtins.foldl'
           81|         (drv: attr: addBuildSystem {
             |               ^
           82|           inherit drv self attr;

       … from call site

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:80:23:

           79|     lib.mapAttrs
           80|       (attr: systems: builtins.foldl'
             |                       ^
           81|         (drv: attr: addBuildSystem {

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/overrides/default.nix:80:14:

           79|     lib.mapAttrs
           80|       (attr: systems: builtins.foldl'
             |              ^
           81|         (drv: attr: addBuildSystem {

       … from call site

       … while evaluating anonymous lambda

       at /nix/store/4yy33bawrm7cmlgmjnjrfn8qvyh0shz8-source/default.nix:43:15:

           42|             (
           43|               dep:
             |               ^
           44|               let

       … from call site

       … while evaluating 'isDerivation'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:427:18:

          426|   */
          427|   isDerivation = x: x.type or null == "derivation";
             |                  ^
          428|

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/stdenv/generic/make-derivation.nix:201:8:

          200|   checkDependencyList' = positions: name: deps: lib.flip lib.imap1 deps (index: dep:
          201|     if lib.isDerivation dep || isNull dep || builtins.typeOf dep == "string" || builtins.typeOf dep == "path" then dep
             |        ^
          202|     else if lib.isList dep then checkDependencyList' ([index] ++ positions) name dep

       … while evaluating anonymous lambda

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/stdenv/generic/make-derivation.nix:200:81:

          199|   checkDependencyList = checkDependencyList' [];
          200|   checkDependencyList' = positions: name: deps: lib.flip lib.imap1 deps (index: dep:
             |                                                                                 ^
          201|     if lib.isDerivation dep || isNull dep || builtins.typeOf dep == "string" || builtins.typeOf dep == "path" then dep

       … from call site

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/lists.nix:117:32:

          116|   */
          117|   imap1 = f: list: genList (n: f (n + 1) (elemAt list n)) (length list);
             |                                ^
          118|

       … while evaluating anonymous lambda

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/lists.nix:117:29:

          116|   */
          117|   imap1 = f: list: genList (n: f (n + 1) (elemAt list n)) (length list);
             |                             ^
          118|

       … from call site

       … while evaluating anonymous lambda

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/stdenv/generic/make-derivation.nix:243:13:

          242|       (map (drv: drv.__spliced.hostHost or drv) (checkDependencyList "depsHostHostPropagated" depsHostHostPropagated))
          243|       (map (drv: drv.crossDrv or drv) (checkDependencyList "propagatedBuildInputs" propagatedBuildInputs))
             |             ^
          244|     ]

       … from call site

       … while evaluating 'getOutput'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/attrsets.nix:598:23:

          597|   */
          598|   getOutput = output: pkg:
             |                       ^
          599|     if ! pkg ? outputSpecified || ! pkg.outputSpecified

       … from call site

       … while evaluating the attribute 'propagatedBuildInputs' of the derivation 'python3.9-market-maker-stats-1.0.0'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/pkgs/stdenv/generic/make-derivation.nix:278:7:

          277|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          278|       name =
             |       ^
          279|         let

       … while evaluating the attribute 'drvPath'

       at /nix/store/8gwzhqn67qm8r5biakx2hxsqiy3jfqmz-source/lib/customisation.nix:210:7:

          209|     in commonAttrs // {
          210|       drvPath = assert condition; drv.drvPath;
             |       ^
          211|       outPath = assert condition; drv.outPath;
```
</details>

K900:
Can you post your code?

asymmetric:
@K900 sorry for spam, I moved this to its own issue.

silky:
fwiw, it still seems like there is an issue here when attempting to _run_ packages, instead of just building an env for them.

i'm in a situation where the nix shell works fine, but `nix run . ...` does not

FRidh:
> fwiw, it still seems like there is an issue here when attempting to _run_ packages, instead of just building an env for them.
> 
> i'm in a situation where the nix shell works fine, but `nix run . ...` does not
> 
> ( -- edit: i think i'm wrong; i solved my problem by adding `setuptools = "*"` as a dep. )

A lot of packages don't add `setuptools` to their runtime dependencies even though they depend on it (typically `pkg_resources`).

silky:
@FRidh what's odd is i certainly dependend on it in test; and also the python that comes in the shell _also_ had `pkg_resources`; it's just the one provided by the `mkPoetryApplication` did not.

silky:
> That's because during build time setuptools may have been available (e.g. because it was used as 
> a build system) but not during run time.

but, being in the shell and running `pytest` (or even `poetry run pytest`) is at "run time", right? ( that's what worked. )

vale981:
> > fwiw, it still seems like there is an issue here when attempting to _run_ packages, instead of just building an env for them.
> > i'm in a situation where the nix shell works fine, but `nix run . ...` does not
> > ( -- edit: i think i'm wrong; i solved my problem by adding `setuptools = "*"` as a dep. )
> 
> A lot of packages don't add `setuptools` to their runtime dependencies even though they depend on it (typically `pkg_resources`).

Is there a workaround for injecting this dependency other than editing `poetry.lock`?

K900:
Yes, via the overrides mechanism.

vale981:
so i feared :P

PR #745, opened 2022-10-06
Various overrides fixes

SemMulder:
Override fixes for `ansible < 5.0.0`, `open3d` and `pyrealsense2`.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -130,9 +130,21 @@ lib.composeManyExtensions [
             self.dopy
             self.ncclient
           ];
-        } // lib.optionalAttrs (lib.versionOlder old.version "5.0") {
-          prePatch = pkgs.python.pkgs.ansible.prePatch or "";
-          postInstall = pkgs.python.pkgs.ansible.postInstall or "";
+        }
+      );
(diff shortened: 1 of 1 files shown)


overrides/default.nix
```diff
@@ -137,9 +137,21 @@ lib.composeManyExtensions [
             self.dopy
             self.ncclient
           ];
-        } // lib.optionalAttrs (lib.versionOlder old.version "5.0") {
-          prePatch = pkgs.python.pkgs.ansible.prePatch or "";
-          postInstall = pkgs.python.pkgs.ansible.postInstall or "";
```
SemMulder, line 142: Note that `prePatch` and `postInstall` have been empty in nixpkgs for a long time :). I looked back in time and added the patches to the right package `ansible-base`.

adisbladis:
I undid the `autoPatchelfHook` `true` -> list change as the API is somewhat broken and doesn't work for us (we can't maintain a static list of ignores but we could have a set of globs).

I've fixed `autoPatchelfHook` in https://github.com/NixOS/nixpkgs/pull/195306.

PR #767, opened 2022-10-12
fix handling of complex build tags

GoldsteinE:


GoldsteinE:
I don’t understand why this failure happened

cpcloud:
@GoldsteinE Sorry it's taken so long to review the PR. Is there a small test you can add that would break without these changes?

adisbladis:
I refactored the code slightly so it could potentially take an arbitrary number of extra dash delimited numbers.
I don't know if this actually happens in the real world, but it seemed prudent.

PR #776, opened 2022-10-19
Add `setuptools` to additional package builds

worldofgeese:
This includes `setuptools` when building `django-cockroachdb` and `psycopg2-binary`.


cpcloud (approved):
LGTM!

PR #778, opened 2022-10-20
test: Verify Black runtime

l0b0:
Does not trigger #775.

### DIFF
--- a/tests/black/default.nix
+++ b/tests/black/default.nix
@@ -1,0 +1,10 @@
+{ poetry2nix, python3, runCommand }:
+let
+  env = poetry2nix.mkPoetryEnv {
+    projectDir = ./.;
+  };
+in
+runCommand "black-test" { } ''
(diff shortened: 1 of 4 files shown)


cpcloud:
@l0b0 Is the idea here to actually trigger #775, or is this to prevent a regression for something that is already fixed elsewhere?

l0b0:
@cpcloud I was hoping to trigger #775, but this should be a valuable regression test nevertheless.

cpcloud:
@l0b0 Does this reproduce if you use `python39` instead of `python3`?

l0b0:
@cpcloud Good call, but it doesn't. At this point, though, I'm moving all my repos to nixpkgs 22.11, which doesn't have this problem.

cpcloud:
Thanks for the PR, but since this isn't reproducing I'm going close it!

PR #779, opened 2022-10-20
fix: eval error on poetry 1.2 when a dev dependency is from git

razielgn:
This fixes an issue I was having when specifying a dev dependency via git. Hopefully this is the right way. The test I have added passed once I did the fix.

The issue was:
```
error: attempt to call something which is not a function but a set

       at /nix/store/fnmpdpj4acacb9ij2sskjj5rdpkp618i-source/default.nix:205:31:

          204|                           or (normalizePackageSet pyProject.tool.poetry.dev-dependencies or { }).${normalizedName}
          205|                           or (normalizePackageSet pyProject.tool.poetry.group.dev.dependencies { }).${normalizedName} # Poetry 1.2.0+
             |                               ^
          206|                           or { }
```


adisbladis:
Thank you!

PR #783, opened 2022-10-25
Add overrides for arpy, pydpkg

mikepurvis:
Without this change, my project that depends on these two packages would only work with `preferWheels = true`.

Would be happy to push these deps upstream as well (or instead), but I'm not really sure how that would work— the sources are here:

- https://github.com/viraptor/arpy/blob/master/setup.py
- https://github.com/memory/python-dpkg/blob/master/pyproject.toml

Honestly it's kind of surprising that `setuptools` especially needs to be explicitly stated as a dep.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -900,6 +900,9 @@
   "arpeggio": [
     "setuptools"
   ],
+  "arpy": [
+    "setuptools"
+  ],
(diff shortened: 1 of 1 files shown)


adisbladis:
Could you add these through the JSON file that adds build systems for other packages?

PR #794, opened 2022-11-04
Fix cryptography `override` not allowing subsequent call to `overridePythonAttrs`

twoolie:
using `.override` creates a derivation that no longer supports `overridePythonAttrs`, so following call to `scrypto.overridePythonAttrs`` fails.

using `overridePythonAttrs` makes the resulting package able to have `overridePythonAttrs` called again.


cpcloud:
This seems reasonable to me. @adisbladis Any thoughts?

PR #795, opened 2022-11-04
overrides.pyopencl: add native OpenCL dependencies and numpy as a build-time dependency

trepetti:


trepetti:
I believe this only failed due to a timeout from a temporarily unavailable resource. Not sure if there is a comment hook to re-run, though.

PR #804, opened 2022-11-07
override: watchfiles 0.18.0

takeda:


takeda:
So with this one, the comment says that `Watchfiles does not include Cargo.lock in tarball released on PyPi for versions up to 0.17.0` and indeed in 0.18.0 the lock file is there, but when trying to use it, it is still complaining that it doesn't have a hash. Isn't it like cryptography or orjson where each version has to have hash, whether there's lock file or not?

PR #808, opened 2022-11-09
poetry2nix: cryptography do not override anything if it is coming from wheel

### LINKED ISSUE #784: overridePythonAttrs missing from scrypto, with preferWheels
I _believe_ this is a new issue with the cryptography overrides in this repo, as my setup was previously working with an older pin of nixpkgs, and manifested once I updated.

Reproduction:
```
$ nix run github:clearpathrobotics/nix-ros-base/with-poetry2nix#nix-generator
error: attribute 'overridePythonAttrs' missing

       at /nix/store/vg80nn2bljqinfag3fqpzngf43sd4dr5-source/overrides/default.nix:384:9:

          383|         in
          384|         scrypto.overridePythonAttrs
             |         ^
          385|           (
```

This is with `preferWheels = true`. Without that, it works if I have https://github.com/nix-community/poetry2nix/pull/783

My `poetry.lock` shows this dependency:
```
[[package]]
name = "cryptography"
version = "37.0.4"
```

However, it does appear that the missing package attribute _is_ in upstream nixpkgs:
```
$ nix eval nixpkgs#python310Packages.cryptography --raw
/nix/store/z1arv7rs6hnm9ks84j1jws83jfmxhpyf-python3.10-cryptography-38.0.1

$ nix eval nixpkgs#python310Packages.cryptography.overridePythonAttrs
<LAMBDA>
```

I'm not sure what the next steps could be here.

vlaci:
My hunch that it is really nothing to do with this package if we have the wheel with everything compiled anyway.

Fixes #784

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -388,37 +388,36 @@ lib.composeManyExtensions [
           );
           sha256 = getCargoHash super.cryptography.version;
           isWheel = lib.hasSuffix ".whl" super.cryptography.src;
-          scrypto =
-            if isWheel then
-              (
-                super.cryptography.override { preferWheel = true; }
-              ) else super.cryptography;
         in
-        scrypto.overridePythonAttrs
+        if isWheel then
           (
-            old: {
-              nativeBuildInputs = (old.nativeBuildInputs or [ ])
-                ++ lib.optional (lib.versionAtLeast old.version "3.4") [ self.setuptools-rust ]
-                ++ lib.optional (!self.isPyPy) pyBuildPackages.cffi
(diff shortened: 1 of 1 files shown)


vlaci:
Interesting. I've glossed over the `propagatedBuildInputs` attribute. I'll check it later in an isolated environment. IMHO won't be an issue as all transitive dependencies would be encoded in the lockfile and explicitly installed by poetry anyway. There shouldn't be a need for any build dependencies to install from a wheel though.

takeda:
I'm running into this same issue when using preferWheels option.

adisbladis:
I think this problem is already fixed by a4bd44e98606a796ea996964f8514f42c073f893.

PR #819, opened 2022-11-19
Add overrides

Nebucatnetzer:
- django-floppyforms
- django-nested-admin
- mixer
- python-monkey-busines
- pytoolconfig

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -3780,6 +3780,9 @@
   "django-filter": [
     "setuptools"
   ],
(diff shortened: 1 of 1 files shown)


Nebucatnetzer:
I know that `python-monkey-business` requires `setuptools` but for some reason the override in the JSON file doesn't seem to be working. But the override in my Flake is working without a problem:
https://github.com/Nebucatnetzer/network_inventory/blob/poetry/flake.nix#L21

PR #820, opened 2022-11-19
Extend docs

Nebucatnetzer:
After battling with the `ModuleNotFoundError` I thought I might try to help with making the docs a bit clearer.
At least the I think that the changes would've helped me.

I was especially confused that you used `poetry` in the example I couldn't' really separate it from `peotry2nix` when debugging the error.
In addition the example used `self.poetry` and in the Matrix chat I was told (and got it working) that I had to use `super.poetry`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ For more information, see [the announcement post on the Tweag blog](https://www.
 
 ## Table of contents
 - [API](#api)
+- [FAQ](#FAQ)
 - [How-to guides](#how-to-guides)
 - [Using the flake](#using-the-flake)
 - [Contributing](#contributing)
@@ -294,6 +295,21 @@ in pkgs.poetry2nix.mkPoetryApplication {
   projectDir = ./.;
 }
 ```
+
+
+## FAQ
+
+**Q.** I'm experiencing one of the following errors, what do I do?
+
+- ModuleNotFoundError: No module named 'setuptools'
+- ModuleNotFoundError: No module named 'pdm'
+- ModuleNotFoundError: No module named 'setuptools-scm'
+- ModuleNotFoundError: No module named 'poetry-core'
+- ModuleNotFoundError: No module named 'flit'
+- ModuleNotFoundError: No module named 'flit-core'
+- ModuleNotFoundError: No module named 'flit-scm'
+
+**A.** Have a look at the following document [edgecase.md](./docs/edgecases.md)
 
 ## How-to guides
 - [Package and deploy Python apps faster with Poetry and Nix](https://www.youtube.com/watch?v=TbIHRHy7_JM)
--- a/docs/edgecases.md
+++ b/docs/edgecases.md
@@ -1,6 +1,8 @@
 # Edge cases
+
 ## Preamble
-The python ecosystem has a very long history.
+
+The Python ecosystem has a very long history.
 Over the years, an uncountable number of modules have been created, but more importantly, the assembly technology of these modules has changed many times.
 `ez_install`, `setuptools`, `distutils`, eggs, `pip`, wheels, PEP-518, PEP-517, `flit`, `poetry`.
 All this diversity creates great difficulties.
@@ -18,47 +20,47 @@ We have tried to collect here and describe typical errors and ways to eliminate 
 
 ## Cases
 
-#### ModuleNotFoundError: No module named 'poetry'
-**Conditions:** you have declared a dependency on git repository with the python package.
-And this package uses poetry as a build tool.
+#### ModuleNotFoundError: No module named 'PACKAGENAME'
+
+**Conditions:** You have declared a dependency in a repository.
+And this package uses `PACKAGENAME` as a build tool. Where `PACKAGENAME` most likely is one of the tools mentioned above (`setuptools`, `pdm`, etc.).
 So since `poetry2nix` cannot obtain this dependency in form of a wheel, it needs to build it from the source by calling pip.
-But pip requires poetry to build this package. And that’s when an error occurs.
+But pip requires `PACKAGENAME` to build this package. And that’s when an error occurs.
 
-**Solution:** in order to make dependency build we need to override it build dependencies by adding poetry  package to it.
+**Solution:** In order to make the dependency build, we need to override its build dependencies by adding the `PACKAGENAME` package to it.
 
-**Example:** let's consider situation when our package has declared `git` dependency in `pyproject.toml` on `extralib`, which uses poetry to be built.
+**Example:** Let's consider the situation when our package has declared `django-floppyforms` as a dependency in `pyproject.toml`, which uses `PACKAGENAME` to be built.
 And we have this declaration in our Nix definition:
+
 ```
 poetry2nix.mkPoetryApplication {
   projectDir = ./.;
 }
 ```
-And trying to build it get us next error log
+
+And trying to build it results in the following error log
 <details>
   <summary>error log (click to expand)</summary>
 
 ```
+Sourcing python-remove-tests-dir-hook
 Sourcing python-catch-conflicts-hook.sh
 Sourcing python-remove-bin-bytecode-hook.sh
-Sourcing pip-build-hook
-Using pipBuildPhase
-Using pipShellHook
 Sourcing pip-install-hook
 Using pipInstallPhase
 Sourcing python-imports-check-hook.sh
 Using pythonImportsCheckPhase
 Sourcing python-namespaces-hook
+Sourcing pip-build-hook
+Using pipBuildPhase
+Using pipShellHook
 @nix { "action": "setPhase", "phase": "unpackPhase" }
 unpacking sources
-unpacking source archive /nix/store/f2mb5sy6vxm81sy5apzvbxmnvj8f62la-source
-source root is source
-setting SOURCE_DATE_EPOCH to timestamp 315619200 of file source/extralib/utils.py
+unpacking source archive /nix/store/w1gk95sf5lknw0mxav5gsvcijcwfqkwh-django-floppyforms-1.9.0.tar.gz
+source root is django-floppyforms-1.9.0
+setting SOURCE_DATE_EPOCH to timestamp 1589942379 of file django-floppyforms-1.9.0/setup.cfg
 @nix { "action": "setPhase", "phase": "patchPhase" }
 patching sources
-Removing path dependencies
-Finished removing path dependencies
-Removing git dependencies
-Finished removing git dependencies
 @nix { "action": "setPhase", "phase": "configurePhase" }
 configuring
 no configure script, doing nothing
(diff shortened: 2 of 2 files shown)


docs/edgecases.md
```diff
@@ -1,49 +1,66 @@
...
 
-All these factors can cause the module not to be assembled or installed in a certain environment. Nix is a much stricter ecosystem, it has more formal requirements, but at the same time more guarantees. And since `poetry2nix` relies on Nix and its strict rules in its work, there are many rough edges and inaccuracies in the metadata on pypi.org leads to errors.
-This is very sad, and we will inevitably face the fact that some modules can not be installed, can not be built, or just cause errors, have unnecessary or missing direct and transitive dependencies. However, Nix is a very flexible system that allows you to modify some aspects of the module with pinpoint accuracy, without changing the source code of the module (and sometimes changing it). This is what this section is dedicated to. We have tried to collect here and describe typical errors and ways to eliminate them.
+The Python ecosystem has a very long history.
```
K900, line 5: Please get rid of the spurious newline changes, they make things difficult to review.

docs/edgecases.md
```diff
@@ -1,49 +1,66 @@
...
 ## Cases
 
-#### ModuleNotFoundError: No module named 'poetry'
-**Conditions:** you have declared a dependency on git repository with the python package. And this package uses poetry as a build tool. So since `poetry2nix` cannot obtain this dependency in form of a wheel, it needs to build it from the source by calling pip. But pip requires poetry to build this package. And that’s when an error occurs.
+#### ModuleNotFoundError: No module named 'PACKAGENAME'
 
-**Solution:** in order to make dependency build we need to override it build dependencies by adding poetry  package to it.
+**Conditions:** You have declared a dependency in a repository.
+And this package uses `PACKAGENAME` as a build tool. Where `PACKAGENAME` most likely is one of the tools mentioned above (`setuptools`, `pdm`, etc.).
+So since `poetry2nix` cannot obtain this dependency in form of a wheel, it needs to build it from the source by calling pip.
```
K900, line 27: poetry2nix prefers sdists to wheels, so this isn't accurate.

docs/edgecases.md
```diff
@@ -52,99 +69,118 @@ building
...
+  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
+  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
+ModuleNotFoundError: No module named 'setuptools'
+
+error (ignored): error: cannot unlink '/tmp/nix-build-python3.10-django-4.1.3.drv-1/Django-4.1.3': Directory not empty
+error: 1 dependencies of derivation '/nix/store/hz4b87s99s1lwiz2m0vwilhlh6rlfj64-python3-3.10.8-env.drv' failed to build
+error: 1 dependencies of derivation '/nix/store/60gaxbdf11lhaj9xg50cf0rr6x5v8v1z-nix-shell-env.drv' failed to build
 ```
+
+As you can see on the fourth last line it's missing `setupools` which in this case is our missing `PACKAGENAME`.
```
K900, line 136: `setuptools`

docs/edgecases.md
```diff
@@ -52,99 +69,118 @@ building
...
+
+``` nix
+poetry2nix.mkPoetryApplication {
+  projectDir = ./.;
+  overrides = poetry2nix.defaultPoetryOverrides.extend
+    (self: super: {
+      django-floppyforms = super.django-floppyforms.overridePythonAttrs
+      (
+        old: {
+          buildInputs = (old.buildInputs or [ ]) ++ [ super.buildtools ];
```
K900, line 150: `setuptools`, again

docs/edgecases.md
```diff
@@ -52,99 +69,118 @@ building
...
+        }
+      );
+    });
+}
 ```
+
+This override will instruct the underlying build logic to include the additional build dependency into the inputs of `django-floppyforms`.
+It might help to know that you don't need to use the full package name like `python39Packages.setupools` but can just use `setuptools` directly.
```
K900, line 158: `setuptools`, again.

docs/edgecases.md
```diff
@@ -52,99 +69,118 @@ building
...
+However you have to use the name like defined in `Nixpkgs` so something like `flit_scm` becomes `flit-scm`.
```
K900, line 159: Not correct, the name should be normalized as per PEP-517. Also, nixpkgs isn't a literal piece of code so shouldn't be in `.`

docs/edgecases.md
```diff
@@ -52,99 +69,118 @@ building
...
+          buildInputs = (old.buildInputs or [ ]) ++ [ super.pdm ];
+        }
+      );
+    });
 }
 ```
-This override will instruct underlying build logic to include additional build dependency into inputs of `extralib`.
+
+We recommend that you contribute your changes to `poetry2nix` so that other users can profit from them as well.
+The file with the upstream overrides can be found here: https://github.com/nix-community/poetry2nix/blob/master/overrides/build-systems.json
```
K900, line 186: Maybe add a few words about how to actually add overrides to that?

K900 (changes requested):

docs/edgecases.md
Nebucatnetzer: Actually it makes it easier to see changes if you put each sentence on a separate line.
That's why I did it as the first commit so that it's clear what I changed afterwards.
In the whole "Files changed" section that might not be so clear but if you look at it commit by commit it makes more sense and future changes will be more obvious.
However, I can still change it if you wish.

docs/edgecases.md
Nebucatnetzer: I didn't write that sentence but IMO it says that no?

It can't obtain wheels so it has to build it and therefore uses sdists?

docs/edgecases.md
Nebucatnetzer: Corrected

docs/edgecases.md
Nebucatnetzer: Corrected

docs/edgecases.md
Nebucatnetzer: Corrected

docs/edgecases.md
Nebucatnetzer: Do you have something specific in mind?

docs/edgecases.md
K900: GitHub handles inline diffs just fine, so I don't think this is necessary.

docs/edgecases.md
K900: It can obtain wheels, it just _chooses not to_.

docs/edgecases.md
K900: Just a simple explanation of the format of the file and where things need to be added.

das-g: That'd be nice, but I don't think it's needed for accepting this PR. The format is IMO quite self-evident from the existing content of that file.

docs/edgecases.md
Nebucatnetzer: New lines are removed.

docs/edgecases.md
Nebucatnetzer: I rewrote the sentence a bit but if it still doesn’t someone with more knowledge about the insides of poetry2nix will have to write it. I’m basically just guessing.

docs/edgecases.md
Nebucatnetzer: I added a sentence about the format.

Nebucatnetzer:
Do you need me to do something else in order to get this merged?

README.md
```diff
@@ -295,6 +296,21 @@ in pkgs.poetry2nix.mkPoetryApplication {
...
+
+- ModuleNotFoundError: No module named 'setuptools'
+- ModuleNotFoundError: No module named 'pdm'
+- ModuleNotFoundError: No module named 'setuptools-scm'
+- ModuleNotFoundError: No module named 'poetry-core'
+- ModuleNotFoundError: No module named 'flit'
+- ModuleNotFoundError: No module named 'flit-core'
+- ModuleNotFoundError: No module named 'flit-scm'
+
+**A.** Have a look at the following document [edgecase.md](./docs/edgecases.md)
```
l0b0, line 312: Would it be better to inline the single edge case here rather than keeping it as a separate document? I feel like it's common enough (a large percentage of Python package updates break because of a missing build tool) that it warrants being more highlighted.

docs/edgecases.md
```diff
@@ -1,49 +1,51 @@
...
 
 ## Cases
 
-#### ModuleNotFoundError: No module named 'poetry'
-**Conditions:** you have declared a dependency on git repository with the python package. And this package uses poetry as a build tool. So since `poetry2nix` cannot obtain this dependency in form of a wheel, it needs to build it from the source by calling pip. But pip requires poetry to build this package. And that’s when an error occurs.
+#### ModuleNotFoundError: No module named 'PACKAGENAME'
 
-**Solution:** in order to make dependency build we need to override it build dependencies by adding poetry  package to it.
+**Conditions:** You have declared a dependency in a repository. And this package uses `PACKAGENAME` as a build tool. Where `PACKAGENAME` most likely is one of the tools mentioned above (`setuptools`, `pdm`, etc.). So since `poetry2nix` prefers to download this dependencies as *sdist*, it needs to build it from the source by calling pip. But pip requires `PACKAGENAME` to build this package. And that’s when the error occurs.
```
l0b0, line 14: Some suggestions:

- Remove reference to "repository"
- "Inline" the `PACKAGENAME` reference
- Expand "sdist"

Maybe something like this:

> **Conditions:** You have declared a Python package dependency. And this package uses one of the build tools mentioned above (`setuptools`, `pdm`, etc.). Since `poetry2nix` prefers to build from source it requires the build tool. And that’s when the error occurs.

docs/edgecases.md
```diff
@@ -52,99 +54,118 @@ building
...
+          buildInputs = (old.buildInputs or [ ]) ++ [ super.pdm ];
+        }
+      );
+    });
 }
 ```
-This override will instruct underlying build logic to include additional build dependency into inputs of `extralib`.
+
+We recommend that you contribute your changes to `poetry2nix` so that other users can profit from them as well.
+The file with the upstream overrides can be found under the following link. It is a simple JSON file which contains the overrides in an array and sorted in alphabetical order. https://github.com/nix-community/poetry2nix/blob/master/overrides/build-systems.json
```
l0b0, line 171: The link could just be inlined, as in

> We recommend that you contribute your changes to [build-systems.json](https://github.com/nix-community/poetry2nix/blob/master/overrides/build-systems.json) so that other users can profit from them as well.

README.md
Nebucatnetzer: IMO the document is too large to be merged into the README and I feel it really requires the detailed explanation.

docs/edgecases.md
Nebucatnetzer: Implemented.

docs/edgecases.md
Nebucatnetzer: Implemented.

cpcloud:
Great work! Thank you!

PR #828, opened 2022-11-24
add setuptools to a few packages that need it

vale981:
see also #736


adisbladis:
Could you add these inputs through `build-systems.json` instead?

vale981:
Thx, wasn't aware of that :].
Will do

vale981:
hmm future is already in there but it didn't work on my box. will check that too

PR #836, opened 2022-12-01
Add more build systems

adisbladis:


l0b0 (approved):
LGTM; which commands do you run to sort the file? Some `jq` thing?

adisbladis:
> LGTM; which commands do you run to sort the file? Some `jq` thing?

`jq --sort-keys`

PR #840, opened 2022-12-03
Add missing setuptools to some django packages

sephii:


overrides/build-systems.json
```diff
@@ -3853,6 +3859,9 @@
   "django-modelcluster": [
     "setuptools"
   ],
+  "django-modeltranslatino": [
```
K900, line 3862: Typo?

sephii: Oops, good catch.

PR #849, opened 2022-12-07
overrides: open3d only include udev on Linux

bouk:
It's not available on macOS


bouk:
@SemMulder

SemMulder:
You might also want to temporarily set `autoPatchelfIgnoreMissingDeps = false;` to check for missing dependencies for macOS specifically.

PR #868, opened 2022-12-11
build: deduplicate nixpkgs import

### LINKED ISSUE #652: Flake creates unecessary instance of nixpkgs when exposing package
At https://github.com/nix-community/poetry2nix/blob/3b01c3e3dc57d511848d8433153ab67db79640e1/flake.nix#L22 an instance of nixpkgs is imported that gets the overlay applied. This costs around 100MB of RAM when evaluating: https://zimbatm.com/notes/1000-instances-of-nixpkgs

From what I can see the overlay is not required because `pkgs.callPackage` can be used directly. If references to the own flake are required `self` can be used.

cpcloud:
Closes #652.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,8 +6,6 @@
 
   outputs = { self, nixpkgs, flake-utils }:
     rec {
-      overlay = import ./overlay.nix;
-
       templates = {
         app = {
           path = ./templates/app;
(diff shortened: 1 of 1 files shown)


cpcloud:
@adisbladis Any thoughts on this one?

adisbladis:
This is wrong as it would cause the `poetry2nix` attribute of `pkgs` to be inherited from nixpkgs, not from this repo.

adisbladis:
Closing in favour of #936

PR #874, opened 2022-12-12
build: speed up pandas build

cpcloud:


cpcloud:
This PR isn't actually necessary.

PR #876, opened 2022-12-12
fix: revert mkdocs-material override

cpcloud:
This partially reverts #865, since the override there doesn't work with `preferWheels = true;`.


cpcloud:
Merging, to get `master` working again.

PR #885, opened 2022-12-15
override: add futurist build system

g00pix:


g00pix:
Hold on I forgot a dependency

g00pix:
Forgot to mention it should be good now.

PR #886, opened 2022-12-15
overrides: build-systems: add poetry-core to hvac

g00pix:
hvac migrated their build system to poetry-core in this commit: https://github.com/hvac/hvac/commit/0ef445367afa4eb0ed5b85fa0d6658cf033f5751

I don't know if we should keep setuptools in the list for compatibility with older versions, let me know what you think.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -6563,7 +6563,14 @@
     "setuptools"
   ],
   "hvac": [
-    "setuptools"
+    {
+      "buildSystem": "setuptools",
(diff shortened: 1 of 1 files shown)


adisbladis:
> I don't know if we should keep setuptools in the list for compatibility with older versions, let me know what you think.

Generally, yes. We still want older versions to work.
Of course if you can marking exactly the versions that needs an override is preferable, but barring that an inflated build closure isn't too bad.

PR #893, opened 2022-12-17
fix: support pyarrow wheels

cpcloud:
Fixes the pyarrow derivation to support wheels. Test included.


cpcloud:
The major motivator here is to allow projects to depend on versions of pyarrow that are released on PyPI but not yet upstreamed in `nixpkgs`.

PR #895, opened 2022-12-17
fix: wheel wheel

cpcloud:
Prevent infinite recursion when trying to use the wheel wheel to wheel-unpack itself.


cpcloud:
@adisbladis Any thoughts on this?

mk-poetry-dep.nix
```diff
@@ -61,6 +61,8 @@ pythonPackages.callPackage
               inherit pythonPackages pyProject;
             } else [ ];
 
+      pname = normalizePackageName name;
+      preferWheel' = preferWheel && pname != "wheel";
```
l0b0, line 65: Is the tick variable idiomatic? It's pretty hard to notice when reading the code, at least. Would it be easier to read as something like `preferWheelExceptSelf` or `resolvedPreferWheel`? (Anyone know a good single word to say "non-self-referencing"?)

adisbladis: It's an FP thing and is read as "prefer wheel prime" and is idiomatic for internal-only variables.

adisbladis (approved):

PR #899, opened 2022-12-17
test: add test for grpcio wheel usage

### LINKED ISSUE #842: `preferWheel = true` ignored for package `panel`
### Describe the issue

Setting `preferWheel = true` in an override for `panel 0.14.1` doesn't result in the wheel been used, even though the package [has a wheel](https://pypi.org/project/panel/0.14.1/#files) and the Python version (3.9) matches.

The package then fails to build, as the build process for build tries to make network requests. 
### Additional context

[This repo](https://github.com/asymmetric/wheel-repro) is a repro test case.

cpcloud:
Adds a test to demonstrate failure to use the grpcio 1.51.1 ~osx 12.0 wheel available on PyPI~ wheel at all.

Fixes #842.


cpcloud:
@adisbladis I could really use your help here, it seems like `preferWheel` has no effect anymore. https://github.com/nix-community/poetry2nix/issues/842 is likely the same issue demonstrated in this PR.

Curiously, throwing `maturin` in as a dependency the `preferWheel` attribute *is* evaluated.

Any idea why this would be the case?

cpcloud:
It seems like reordering the overlays in `overrides.withDefaults` put the user overlay before the default poetry overrides overlay achieves the desired behavior. Admittedly, this is not a solution that I arrived at from thorough analysis.

My rationale was that it seemed like eval isn't happening at all on many packages, so I tried "forcing" eval by putting user overrides first in the list of overlays.

cpcloud:
@asymmetric Can you try this PR out on your `panel` use case? I've incorporated a test for it here, and I think I've addressed the issue of `preferWheel` not ever being evaluated.

asymmetric:
@cpcloud thanks for looking into this!

We actually don't use `withDefaults, we do

```nix
      mkPoetryApplication = {
        projectDir = ./.;
        python = pythonVersion;
        overrides = [
          pkgs.poetry2nix.defaultPoetryOverrides
          (final: prev: {
            panel = prev.panel.override { preferWheel = true; };
            bokeh = prev.bokeh.override { preferWheel = true; };
          })
          # more custom overlays
        ];
      };
```

I moved the inline overlay above `pkgs.poetry2nix.defaultPoetryOverrides`, and it indeed solves the issue!

One thing I don't understand though is why the `-A preferWheel` test passes even without this PR.

tests/grpcio-wheel/default.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, poetry2nix, python3, pkgs, runCommand }:
+let
+  inherit (pkgs.stdenv) isLinux;
+  env = poetry2nix.mkPoetryEnv {
+    python = python3;
+    pyproject = ./pyproject.toml;
+    poetrylock = ./poetry.lock;
+    preferWheels = false;
```
l0b0, line 8: Are any of the above needed? They are all default values, right? (Let me know if this sort of comment is too nitpicky; as a developer I'm used to getting rid of any duplication of defaults.)

cpcloud:
@adisbladis Any chance you can take a look at this PR?

tests/grpcio-no-wheel/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.poetry]
+name = "test-grpcio-wheel"
```
asymmetric, line 2:
```suggestion
name = "test-grpcio-no-wheel"
```

tests/grpcio-no-wheel/pyproject.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+description = "Test of wheel for grpcio"
```
asymmetric, line 4:
```suggestion
description = "Test of sdist for grpcio"
```

tests/grpcio-wheel/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.poetry]
```
asymmetric, line 1: Should we share the `pyproject.toml`/`.lock` files across wheel/no-wheel, to reduce chances that they diverge over time?

tests/grpcio-no-wheel/default.nix
```diff
@@ -0,0 +1,22 @@
+{ lib, poetry2nix, python3, runCommand }:
```
asymmetric, line 1: Do we need the no-wheel test, since it's the default behaviour of poetry2nix to not install wheels?

What is being tested in this case? If there's anything special about grpcio, it would be nice to mention it at the top of this file.

tests/grpcio-wheel/default.nix
```diff
@@ -0,0 +1,21 @@
...
+        grpcio = super.grpcio.override {
+          preferWheel = isLinux;
+        };
+      }
+    );
+  };
+  isWheelGrpcIO = env.python.pkgs.grpcio.src.isWheel;
+in
+assert isLinux -> isWheelGrpcIO; runCommand "grpcio-wheel" { } ''
+  ${env}/bin/python -c 'import grpc; print(grpc.__version__)' > $out
```
asymmetric, line 20: What's the advantage of evaluating to this rathern than just `env`?

asymmetric:
> One thing I don't understand though is why the -A preferWheel test passes even without this PR.

@cpcloud do you have any ideas about this?

tests/grpcio-no-wheel/default.nix
cpcloud: I think I was trying to be thorough here, but maybe it's better to just wait until there's an actual thing not working to add a test?

tests/grpcio-wheel/default.nix
cpcloud: Nothing I guess. I wanted to test the import, but `env` is probably sufficient.

tests/grpcio-wheel/pyproject.toml
cpcloud: That sounds good, but I'll just remove the `no-wheel` test for now.

tests/grpcio-wheel/default.nix
```diff
@@ -0,0 +1,18 @@
+{ lib, poetry2nix, python3, pkgs, runCommand }:
```
asymmetric, line 1: One more thing: what does this test catch that the panel/bokeh one doesn't?

Could we just limit ourselves to one of the two? If not, then maybe adding an explanation of what exactly is being tested could be helpful.

cpcloud: Sure, grpcio fails because of a missing wheel on the latest GitHub macos runners.

yorickvP:
Applying defaults after user overlays feels wrong to me, and it does in fact break my code. It prevents users from undoing changes made in the default overlays.
If `preferWheel` doesn't compose right that's a deeper issue that we should fix.

Please revert this.

Code that breaks:
```
poetry2nix.overrides.withDefaults (super: self: {
  pytest = null;
})
```

cpcloud:
@adisbladis @yorickvP's request seems reasonable. Shall we revert?

adisbladis:
I'm OK with a revert.

asymmetric:
Any ideas about what the underlying cause of the lack of composability of `preferWheel`?

yorickvP:
Yes, `.override` is from the callPackage call in `mk-poetry-dep.nix`, while `overridePythonAttrs` comes from its call to `buildPythonPackage`, which doesn't know about the outer `callPackage`.

Options:
- override `overridePythonAttrs` to know about the outer `callPackage`.
- change buildPythonPackage to take a function, just like `mkDerivation` these days, and specify `preferWheel` through `overridePythonAttrs`
- remove `preferWheel` somehow
- ???

iwanb:
> Applying defaults after user overlays feels wrong to me, and it does in fact break my code. It prevents users from undoing changes made in the default overlays. If `preferWheel` doesn't compose right that's a deeper issue that we should fix.
> 
> Please revert this.
> 
> Code that breaks:
> 
> ```
> poetry2nix.overrides.withDefaults (super: self: {
>   pytest = null;
> })
> ```

I encountered the same issue. Will the change be reverted? As a workaround I did:
```
overrides = [ poetry2nix.defaultPoetryOverrides  ...];
```

sigprof:
This also breaks in cases like #1519, or when trying to apply a local override for `rpds-py` while #1502 is still pending.  The workaround in https://github.com/nix-community/poetry2nix/pull/899#issuecomment-1620306977 helps, but why then have `overrides.withDefaults` if it can't be used sanely?

PR #918, opened 2023-01-03
feat: Show lock location in package errors

roberth:
Tested by temporarily setting `meta.broken = true;` in `mk-poetry-dep.nix`, and then

    $ nix-build tools -A env
    error: Package ‘python3.9-click-8.1.3’ in /home/user/h/poetry2nix/tools/poetry.lock:0 is marked as broken, refusing to evaluate.

       a) To temporarily allo[...]

As you can see, the location of the lock file is reported, which is especially useful in Nixpkgs, which contains multiple lockfiles.

This used to be just the location of a generic python builder, which is generally not what the user wants.

I've exposed the `poetrylockPos` variable as a parameter that callers may set. This is intended as a contingency more so than a feature that is useful to users. That's why I've decided not to document it, because it would just add noise. I could make it internal, a let binding, if you prefer.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -137,6 +137,7 @@ lib.makeScope pkgs.newScope (self: {
     { projectDir ? null
     , pyproject ? projectDir + "/pyproject.toml"
     , poetrylock ? projectDir + "/poetry.lock"
+    , poetrylockPos ? { file = toString poetrylock; line = 0; column = 0; }
     , overrides ? self.defaultPoetryOverrides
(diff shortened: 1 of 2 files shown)


adisbladis:
Sweet! This is a fantastic addition!

adisbladis:
> I could make it internal, a let binding, if you prefer.

I think leaving it undocumented is OK and we can simply consider it a "private API".

PR #920, opened 2023-01-03
fix: Use hatchling for filelock >=3.9.0

l0b0:
As per the [release log](https://github.com/tox-dev/py-filelock/releases/tag/3.9.0).

Should unblock https://github.com/linz/geostore/pull/2548.


cpcloud (approved):
Thank you!

PR #923, opened 2023-01-06
Update filelock build systems

utybo:
Hi!

`filelock` has [recently](https://github.com/tox-dev/py-filelock/commit/1cf8fcae4e688c60a078b1ec9286637141fb35bf#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R2-R3) switched to `hatchling` + `hatch-vcs`. This PR updates the corresponding override. I've kept setuptools and setuptools-scm in order to keep compatiblity with older versions (not sure if this is the recommended pattern for this).

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -4955,7 +4955,9 @@
   ],
   "filelock": [
     "setuptools",
-    "setuptools-scm"
+    "setuptools-scm",
+    "hatchling",
+    "hatch-vcs"
   ],
   "filemagic": [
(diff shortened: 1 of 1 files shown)


adisbladis:
There is a thing for setting version bounds on build inputs, but inflated build closures are not a big deal imho so this is fine.
If someone wants to implement more specific bounds to get smaller build closures I'll happily merge a follow-up with that.

PR #925, opened 2023-01-06
Improve issue template: add syntax highlighting, don't make a bullet list

das-g:
- don't make a bullet list in section "Additional context"

  A bullet list would make sense if the code blocks would be part of the items. To make a code block part of a list item in Markdown, it would though have to be indentedwhich is either too laborious (when indenting all lines) or too confusing (when indenting only the opening line). If they aren't indented, they'd interrupt the list (thus effectively splitting it into 3 one-item lists) so let's instead just have the file names as prose-ish captions of the code blocks.

- add syntax highlighting

  We do know what languages we expect in the files we ask the issue author to provide in the "Additional context" section, so let's set up proper syntax highlighting for them.


cpcloud (approved):
LGTM. Thanks!

PR #946, opened 2023-01-17
fix: wheel url parsing

ryanswrt:
Currently (as far as I can tell) all url dependency specs are treated as source tarballs, and specifying wheel urls fails. 
I need to add wheels manually from a non pypi source (https://data.pyg.org/whl/) which gave an error when trying to build it -  
a) didn't use the hash of the wheel, so I had to use --impure and 
b) thought that the wheel is a tarball, not a zip
c) didn't know how to use the wheel format if the url points to a wheel

This PR addresses those issues

### DIFF
--- a/mk-poetry-dep.nix
+++ b/mk-poetry-dep.nix
@@ -185,6 +185,7 @@ pythonPackages.callPackage
           builtins.fetchTarball
             {
               inherit (source) url;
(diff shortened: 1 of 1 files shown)


cpcloud:
LGTM, thanks!

PR #951, opened 2023-01-20
feat: Configure Dependabot for all poetry.lock files

### LINKED ISSUE #932: RFC: Create dependency update PRs automatically?
There are currently 70 pyproject.toml files in this repo, with 1060 package references in the lock files. Would it be interesting to automatically create pull requests to update these package references? This would make it much easier to ensure that poetry2nix keeps working with newer versions of Python packages. This way new releases which fail to build within poetry2nix could be caught really quickly, and fixed by adding the relevant commit to the existing PR. Of course, some exceptions might be necessary to keep references to old packages which need to be tested differently from new packages. I've used Dependabot and Renovate for many repos, and it's been really helpful. I can provide a PR for this change if you'd like.

l0b0:
Some of these will probably have to be tweaked because they are tailored to specific versions of a package. This can be done by [ignoring](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) the relevant package. That way, the test can still get updated versions of any other packages, so it stays relevant as long as possible. An alternative approach is to not ignore anything, but to create a new test whenever a Dependabot update breaks an existing test.

Other thoughts:

- This will result in a large number of Dependabot PRs, at least at first. It might be worth upping or removing the [limit on open PRs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit), at least to begin with.
- Some packages might be updated very often, like semi-daily, and could create a fair bit of overhead to just deal with the PRs. Two options:
   - Reduce the update frequency to weekly or monthly, at the risk of missing a broken package for a long-ish time. Not recommended since there are probably many poetry2nix users who update on a daily basis.
   - Configure Kodiak to [merge passing Dependabot PRs automatically](https://kodiakhq.com/docs/config-reference#mergeautomerge_dependenciesversions).

Generated with the following code:

```bash
git ls-files '**/poetry.lock' | xargs dirname | sort | while read dir; do printf '  - package-ecosystem: "pip"
    directory: "/%s"
    schedule:
      interval: "daily"
' "$dir"; done >> .github/dependabot.yml 
```

Closes #932.

(I might be unavailable for comment until ~January 28th.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
     directory: "/pkgs"
     schedule:
       interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tools"
+    schedule:
+      interval: "daily"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,287 @@
 version: 2
 updates:
-  # Auto-update the Poetry package, but not tests
   - package-ecosystem: "pip"
     directory: "/pkgs"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/aiopath"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/ansible-molecule"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/assorted-pkgs"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/awscli"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/bcrypt"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/tests/canonical-module-names"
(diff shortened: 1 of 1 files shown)


cpcloud:
Can we just throw `/tests` at it and avoid having to remember to add a new directory to this file every time a test is added?

cpcloud:
Another option is to use mend renovate which has plenty of configuration options to deal with the issues you outline here.

l0b0:
> Can we just throw `/tests` at it and avoid having to remember to add a new directory to this file every time a test is added?

[Dependabot doesn't seem to support recursive detection of dependencies yet.](https://github.com/dependabot/dependabot-core/issues/1015)

> Another option is to use mend renovate which has plenty of configuration options to deal with the issues you outline here.

Renovate is great - with an absolutely [minimal configuration](https://gitlab.com/engmark/root/-/blob/bae891911353b58530fcde0c078a2854e74b46b5/renovate.json) it recursively detects all types of dependencies. The only downside is having to create a job to trigger it.

l0b0:
@adisbladis What do you think?

cpcloud:
You don't need a job to trigger renovate. It runs on a schedule.

l0b0:
OK, I'll configure that instead then. Cheers!

PR #961, opened 2023-01-25
Add setuptools build dependency to ligo-segments

danielunderwood:
I'm not sure if there's a typical format for these PRs, but this adds setuptools as a build dependency to the ligo-segments package to fix builds that depend on that package.


cpcloud:
This is typical format :)

Merging! Thank you.

PR #976, opened 2023-02-03
overrides: automat: only add m2r to inputs for older versions

delroth:
Newer versions of automat do not depend on m2r anymore [1]. Because of this m2r might not end up being in poetry.lock along with automat, so "self.m2r" ends up refering to the nixpkgs pythonPackages.m2r. That package happens to be marked as broken in some conditions.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -99,7 +99,7 @@ lib.composeManyExtensions [
 
     {
       automat = super.automat.overridePythonAttrs (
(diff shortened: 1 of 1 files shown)


l0b0 (approved):
Confirmed removed in https://github.com/glyph/automat/commit/1a687e23fef2895c8acf2f012d77a0f598d28642; LGTM!

PR #977, opened 2023-02-03
overrides: add libarchive

delroth:
- Build system dependency (setuptools).
- Native library dependency path fixup.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1002,6 +1002,15 @@ lib.composeManyExtensions [
         }
       );
 
+      libarchive = super.libarchive.overridePythonAttrs (old: {
+        buildInputs = (old.buildInputs or [ ]) ++ [ self.setuptools ];
+
+        postPatch = ''
+          substituteInPlace libarchive/library.py --replace \
+            "_FILEPATH = find_and_load_library()" "_FILEPATH = '${pkgs.libarchive.lib}/lib/libarchive${stdenv.hostPlatform.extensions.sharedLibrary}'"
+        '';
+      });
+
       libvirt-python = super.libvirt-python.overridePythonAttrs ({ nativeBuildInputs ? [ ], ... }: {
         nativeBuildInputs = nativeBuildInputs ++ [ pkg-config ];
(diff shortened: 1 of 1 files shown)


overrides/default.nix
```diff
@@ -1002,6 +1002,15 @@ lib.composeManyExtensions [
         }
       );
 
+      libarchive = super.libarchive.overridePythonAttrs (old: {
+        propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ self.setuptools ];
```
l0b0, line 1006: Does this need to be propagated? Otherwise, the current idiomatic way to do this is a

```json
"libarchive": [
  "setuptools"
]
```

entry in `overrides/build-systems.json`.

delroth: Let me double check actually -- I think I originally had it propagated because I was debugging some issue with the module, but in the end I can't remember if it was actually required or not, and it seems weird for this module to have a runtime dep on setuptools...

Since the `postPatch` is required either way would you rather have two separate entries for this module (one here, one in `build-systems.json`) or just one entry here? I see a few other entries in `default.nix` that add build system info along with other overrides (e.g. `datadog-lambda`, `molecule`).

l0b0: I'm not sure whether there's an actual policy, I guess the main draw of putting it in `build-systems.json` would be less code overall. @adisbladis @cpcloud?

delroth: Had some time to do more testing and indeed the propagation was not needed. Thanks for catching this!

Leaving the build system override in default.nix for now until further input. I don't care strongly either way, but I think that keeping the overrides in one place is slightly better, and it also means only one layer of overriding instead of two which should be marginally more efficient? Let me know if you or anyone feels strongly about splitting it out into a build-systems.json entry.

adisbladis: > I'm not sure whether there's an actual policy, I guess the main draw of putting it in build-systems.json would be less code overall. @adisbladis @cpcloud?

The intent with `build-systems.json` was to make the simple 99% of cases where a build system is missing trivial, and that anything that didn't fit this very simple use case were to go in the Nix expression.

PR #979, opened 2023-02-06
build-systems: add bencode-py

vamega:
I'm unable to explain why the CI is failing.
The parse error happens on a line that doesn't seem related to the lines I'm changing here.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -1873,6 +1873,9 @@
   "bellows": [
     "setuptools"
   ],
+  "bencode-py": [
+    "setuptools"
+  ],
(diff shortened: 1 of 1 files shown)


overrides/build-systems.json
```diff
@@ -1873,6 +1873,9 @@
   "bellows": [
     "setuptools"
   ],
+  "bencode-py" [
```
l0b0, line 1876:
```suggestion
  "bencode-py": [
```

vamega: Wow, I was completely blind when doing this and missed the obvious.

PR #985, opened 2023-02-12
build: Support virtualenv 20.18 using hatch-vcs

l0b0:


adisbladis:
Did you leave out cython by accident or is it not really required?

l0b0:
> Did you leave out cython by accident or is it not really required?

virtualenv 20.17.1 and 20.18.0 both build without cython, so I thought I'd just leave it out.

PR #996, opened 2023-02-24
build: Bump nixpkgs

l0b0:
With `niv update`, using the unstable branch.

Requires some workaround for

> SyntaxError: future feature annotations is not defined

(maybe https://github.com/NixOS/nixpkgs/pull/218399?) and #1014 to avoid

> error: attribute 'TESTING_FOOBAR' missing


adisbladis:
Continued in https://github.com/nix-community/poetry2nix/pull/1034

PR #1011, opened 2023-02-27
Bump nixpkgs

adisbladis:


adisbladis:
Continued in https://github.com/nix-community/poetry2nix/pull/1034

PR #1012, opened 2023-02-27
chore: Remove nixops references

### LINKED ISSUE #1010: Remove nixops dependencies?
### Describe the issue

While trying to debug why #996 is failing to build I came across a couple comments (like [this one](https://github.com/NixOS/nixpkgs/issues/217526#issuecomment-1442835731)) which indicate that nixops should be considered deprecated. Would it be possible to simply remove any references to it?

l0b0:
Closes #1010.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -59,15 +59,6 @@ let
 
 in
 lib.composeManyExtensions [
-  # NixOps
-  (self: super:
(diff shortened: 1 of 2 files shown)


l0b0:
OK, we'll need some other workaround for the

> SyntaxError: future feature annotations is not defined

when using newer nixpkgs then.

PR #1013, opened 2023-02-28
fix: Use new `propagatedBuildInputs`

l0b0:
As per latest nixpkgs warning:

> 'deps' argument to makeSetupHook is deprecated and will be removed in
> release 23.11.


l0b0:
Can't be merged without also [bumping nixpkgs](https://github.com/nix-community/poetry2nix/pull/996).

PR #1027, opened 2023-03-02
Add overrides for hikari and hikari-lightbulb

Lunarequest:
`hikari` is a library to write discord bots in python and `hikari-lightbulb` is a command handler for `hikari`. Both of these can be used in poetry2nix with some minor tweaks


Lunarequest:
cachix seems to be broken :/

adisbladis:
We've fixed the CI action! Looks good!

PR #1033, opened 2023-03-05
feat: orjson 3.8.7

ac130kz:


adisbladis:
I can't force-push to your branch to rebase this PR, but the hash has been manually verified so it's OK.

PR #1051, opened 2023-03-12
fix: gitlint-core 0.19 needs hatchling

l0b0:
Not sure how to fix this - adding `hatch-vcs` does not fix the build issue.


Pegasust:
Could you include the error in the build log too?

adisbladis:
It also needed hatch-vcs.

PR #1061, opened 2023-03-16
Add `option` and `returns` python module (for `option` and `result` in Python)

Pegasust:


tests/fastapi-utils/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.poetry]
+name = "test-fastapi-utils"
+version = "0.1.0"
+description = "Test of python fastapi-utils package"
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = "^3.6"
```
l0b0, line 8: [3.7 is the oldest supported version](https://www.python.org/downloads/); does it make sense to limit to that instead?

adisbladis: Yes, that sounds good.

adisbladis: In the mean time I don't think it matters much.

PR #1064, opened 2023-03-16
Add build system for fastapi-utils

Pegasust:
- [x] Add new entry to `overrides/build-systems.json`
- [x] Add simple test to validate successful install (`tests/fastapi-utils/*`)
- [x] Run locally to validate successful build


adisbladis:
Could you remove the test? I don't think it's adding any real value and test runs already take way too much time.

PR #1082, opened 2023-03-25
fix: fix `and`/`or` precedence in marker parsing

cpcloud:
~This PR adds a failing test for sqlalchemy 1.4.47 using poetry 1.4. I am not sure what the issue is yet. cc @adisbladis, any idea what might be going wrong here?~

`and`/`or` were being treated the same in operator precedence marking, this PR fixes the issue


cpcloud:
It looks like the generated wheel, which is

```
SQLAlchemy-1.4.47-cp310-cp310-linux_x86_64.whl
```

doesn't match any of the wheels in the lock file, causing the dependency to not be found.

This doesn't show up when depending on SQLAlchemy 2.0, so I'm look upstream to see if there's a patch to bring in.

cpcloud:
Ok, that doesn't seem related. SQLAlchemy 2.0 generates a similarly named wheel.

cpcloud:
I think this might be a python marker parsing issue 😅

cpcloud:
This going to need a more thorough solution. I'm working on it.

cpcloud:
Ok, I'm not entirely sure how to proceed. 

Here's the issue:

1. Markers without parentheses indicating precedence are treated without precedence, so `and`s are not necessarily evaluated before `or`s
2. Ideally we can use the `packaging` module's builtin support for evaluating markers, but it seems that because of cross compilation we can't necessarily use the target platform's python interpreter to evaluate the markers
3. `packaging` is only python 3 compatible, so any support for python2 would have to be dropped. I get a Python 2 related failure when trying to run all the tests locally.

adisbladis:
> Ideally we can use the packaging module's builtin support for evaluating markers, but it seems that because of cross compilation we can't necessarily use the target platform's python interpreter to evaluate the markers

That would mean IFD so it's never going to work. There is a reason why the marker evaluation was implemented in Nix.

PR #1085, opened 2023-03-27
fix: Remove Unicode BOM in mutmut's setup.py

### LINKED ISSUE #1084: "invalid non-printable character U+FEFF" in mutmut using poetry2nix 1.40.1
### Describe the issue

When attempting to upgrade from poetry2nix 1.39.1 to 1.40.1 I can no longer install mutmut.

### Additional context

Build log:

```text
$ nix log /nix/store/nnc2cn96rnlj41iwnf73skw3msxi9x8d-python3.10-mutmut-2.4.3.drv
Sourcing pip-install-hook
Using pipInstallPhase
Using pythonImportsCheckPhase
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
unpacking sources
unpacking source archive /nix/store/wf4cxw6bczmp9mhi94dyb03jrqlmgw5i-mutmut-2.4.3.tar.gz
source root is mutmut-2.4.3
setting SOURCE_DATE_EPOCH to timestamp 1670777864 of file mutmut-2.4.3/setup.cfg
patching sources
  File "/nix/store/nry474aicns0vk6f1y76via6g1snj8ms-python-requires-patch-hook.py", line 70, >
    tree = ast.parse(f.read())
  File "/nix/store/0n4y44dnaxafqs7cg625aldrb152x7bx-python3-3.10.10/lib/python3.10/ast.py", l>
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    ﻿#!/usr/bin/env python3
    ^
SyntaxError: invalid non-printable character U+FEFF
```

~presumably this is as simple as a patch to remove the offending character.~

Based on [discussion with mutmut's author](https://github.com/boxed/mutmut/issues/276) I think the fault lies with the poetry2nix Python file parsing using `str` rather than `bytes`. We should probably change     `hooks/python-requires-patch-hook.py` and related code to read/write Python files as bytes rather than strings.

l0b0:
I'm not sure if there's a better way to represent this in Nix, because the docs don't mention any way to escape bytes/Unicode <https://nixos.org/manual/nix/stable/language/values.html#type-string>, `\uEFBBBF` and `\xEF\xBB\xBF` don't seem to work, and Unicode support was rejected back in 2018 <https://github.com/NixOS/nix/issues/770>. I had to paste the result of `head --bytes=3 setup.py | xclip` verbatim, which is why it shows up as `ZWNBSP` in JetBrains IDEs and an empty string in `git diff`.

😢

To detect BOM in files:

1. Unpack the package tarball
2. Go to the extracted directory
3. `git init`
4. `git add .`
5. `dos2unix *`
6. Check which files changed (some of them might just have had Windows newlines)

Closes #1084.


adisbladis:
Maybe we should deal with this in a hook instead? That way any other problematic packages would be dealt with automatically.

l0b0:
> Maybe we should deal with this in a hook instead? That way any other problematic packages would be dealt with automatically.

Do we want the risk/computational cost of dealing with it automatically, though? This looks like the first case we've had to deal with so far, and it's one of very few instances I've ever seen of an UTF-8 BOM.

adisbladis:
> Do we want the risk/computational cost of dealing with it automatically, though? This looks like the first case we've had to deal with so far, and it's one of very few instances I've ever seen of an UTF-8 BOM.

I actually think the fix is as simple as just assuming that all `setup.py` files are UTF-8 with BOM and open them as such.
Calling `open()` with `encoding='utf-8-sig'` makes no difference for files without a BOM but makes a difference for files with one.

Issue #672, opened 2022-06-20
How to use plugin system, trying to use keyrings-google-artifactregistry-auth


DylanRJohnston-FZ:
### Describe the issue
I'm trying to make use of poetry2nix where we have a private artifact repository on google cloud. Normally the keyring plugin keyrings.google-artifactregistry-auth allows Poetry to authenticate against the private repository. However it's unclear how to achieve this with poetry2nix.

### Additional context
I initially tried using an overlay to put `keyrings-google-artifactregistry-auth` in the `propogatedBuildInputs` but quickly realised this doesn't work because Poetry doesn't use the typical nixpkgs callPackage convention and instead uses poetry itself to manage its own dependencies.

I could potentially achieve this by overriding the pyproject and poetrylock files to include the google keyring plugin.

```
, pyproject ? projectDir + "/pyproject.toml"
, poetrylock ? projectDir + "/poetry.lock"
```

However when searching the repository, I found an undocumented `withPlugins` method that I was hoping you'd help me figure out how to use.

DylanRJohnston-FZ:
poetry.toml
```toml
[repositories]
[repositories.redacted]
url = "https://redacted.pkg.dev/redacted/python"
```
pyproject.toml
```toml
[[tool.poetry.source]]
name = "redacted"
secondary = true
url = "https://redacted.pkg.dev/redacted/python/simple/"
```

rszamszur:
I guess `withPlugins` is experimental/not yet implemented.

The only idea I have at the moment is try to build `poetry2nix` with custom `poetry`:
https://github.com/nix-community/poetry2nix/blob/3cc82ca6ccca2a06dbb05cb0b74efd5d50e60d48/default.nix#L3
such that would already contain the required plugin. However, I do not know if it would work. Might give a try in spare time.

MatrixManAtYrService:
The word plugin is often overloaded...  Are we talking about [poetry plugins](https://python-poetry.org/docs/master/plugins/#the-self-add-command) like you'd put here? 

```poetry self add myplugin```

If so, maybe a good step would be documenting what is and isn't supported re: plugins in the readme?  For instance, my team calls [this plugin](https://pypi.org/project/poetry-plugin-sort/) in a pre-commit hook and I'm not sure if adding it the plugin to the poetry in my devshell is unsupported, or if I'm just overlooking the way to do it.

DylanRJohnston-FZ:
[...]

PR #1106, opened 2023-04-05
fix(flake): build poetry with correct poetry2nix

K900:
Also remove the completely unused `poetry` argument from `default.nix` as to not confuse people even more.


DylanRJohnston-FZ:
What is the correct approach to overriding the `poetry` package now that the argument is removed? I need to add the `keyrings-google-artifactregistry-auth` package to poetry. As per #672

PR #1107, opened 2023-04-05
More overrides-leaking-around fixes

K900:


overlay.nix
```diff
@@ -1,7 +1,7 @@
-final: prev: {
+final: prev: rec {
 
   poetry2nix = import ./default.nix { pkgs = final; };
 
-  poetry = prev.callPackage ./pkgs/poetry { python = final.python3; };
+  poetry = prev.callPackage ./pkgs/poetry { python = final.python3; inherit poetry2nix; };
```
adisbladis, line 5: Using `rec` in overlays have some problematic scoping implications.
It's better to use `inherit (final) poetry2nix;`.

adisbladis (changes requested):

overlay.nix
K900: Good call, applied.

PR #1109, opened 2023-04-07
Fix for torch 2.0 (with CUDA 11)

sterlind:
Torch 2.0 recently became stabilized and became the default. Torch has changed the way they package their binaries recently:

- Prior to 1.13.0, the whole enchilada of cudnn, cufft etc. was bundled inside the massive 1.16GB Torch wheel.
- As of 1.13.0, there's explicit dependencies on CUDA pip packages (e.g. `nvidia-cuda-runtime-cu11`.)
- As of 2.0.0, these have multiplied to a smörgåsbord of CUDA dependencies.
- torch 2.0.0 also depends on triton, which depends on torch! circular dependencies, neat!

It looks like the original torch overrides here inject `cudatoolkit` from nixpkgs, but this looks extremely fragile. The Torch team doesn't really support using system CUDA, unless you build Torch yourself.

Anyway, there were enough changes that I rewrote the override, more or less. I'm very new to nix, please let me know if this is okay!


sterlind:
@adisbladis did you see this PR? is there anything I can do to help get it merged?

tfmoraes:
@sterlind I'm having this problem when I try to use `torch` with `cuda`:

```
Could not load library libcudnn_ops_infer.so.8. Error: libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory
```

`libcudnn_ops_infer.so.8` is contained on `nvidia-cudnn-cu11`.

PR #1124, opened 2023-04-21
Add overrides for `insights-core` and `pynetbox`

KiruyaMomochi:
- overrides: add insights-core
- overrides: add pynetbox

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1946,6 +1946,10 @@ lib.composeManyExtensions [
         }
       );
 
+      pynetbox = super.pynetbox.overridePythonAttrs (old: {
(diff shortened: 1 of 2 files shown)


overrides/default.nix
```diff
@@ -1946,6 +1946,10 @@ lib.composeManyExtensions [
         }
       );
 
+      pynetbox = super.pynetbox.overridePythonAttrs (old: {
```
adisbladis, line 1949: Why isn't this added in the build systems json file?

cpcloud: @KiruyaMomochi Any thoughts here?

KiruyaMomochi: Sorry I think I have sent the comment but it have stayed in pending state for one week...

cpcloud: Ah, I guess I missed that!

PR #1140, opened 2023-04-26
overrides: Add overrides for several packages

PicoGeyer:
Add overrides for
cysystemd
pem
pip-licenses
python-pam
pythonping
tacacs-plus

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -432,6 +432,11 @@ lib.composeManyExtensions [
         '';
       });
 
+      cysystemd = super.cysystemd.overridePythonAttrs (old: {
+        buildInputs = (old.buildInputs or [ ]) ++ [ super.setuptools pkgs.systemd ];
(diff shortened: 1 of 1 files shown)


overrides/default.nix
```diff
@@ -2060,6 +2074,11 @@ lib.composeManyExtensions [
         '';
       });
 
+
+      pythonping = super.pythonping.overridePythonAttrs (old: {
```
adisbladis, line 2078: Could you do this in build-systems.json? (same goes for other instances of adding build systems).

PicoGeyer: Thanks, I'll update the pull request.

PicoGeyer: I apologize for the mess, but I somehow made a new pull request instead of updating this one. So my changes are on #1144. I'll close this one out.

PR #1144, opened 2023-04-28
Various module overrides

PicoGeyer:
Add overrides for
cysystemd
pem
pip-licenses
python-pam
pythonping
tacacs-plus

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -432,6 +432,11 @@ lib.composeManyExtensions [
         '';
       });
 
+      cysystemd = super.cysystemd.overridePythonAttrs (old: {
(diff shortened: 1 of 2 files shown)


overrides/default.nix
```diff
@@ -432,6 +432,11 @@ lib.composeManyExtensions [
         '';
       });
 
+      cysystemd = super.cysystemd.overridePythonAttrs (old: {
+        buildInputs = (old.buildInputs or [ ]) ++ [ super.setuptools pkgs.systemd ];
```
adisbladis, line 436: Could you add setuptools through the json file?

PicoGeyer: Okay, sure. I wasn't sure what was preferred for that case.

PR #1153, opened 2023-05-04
Switch jsonref to pdm

Goorzhel:
https://github.com/gazpachoking/jsonref/commit/9faf61eb8

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -7941,7 +7941,14 @@
     "setuptools"
   ],
   "jsonref": [
-    "setuptools"
+    {
+      "buildSystem": "setuptools",
+      "until": "1.1.0"
+    },
+    {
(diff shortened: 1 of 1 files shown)


adisbladis:
Could you add a version range guard to the change so older builds still work?

overrides/build-systems.json
```diff
@@ -7941,7 +7941,14 @@
     "setuptools"
   ],
   "jsonref": [
-    "setuptools"
+    {
+      "buildSystem": "setuptools",
+      "until": "v1.1.0"
```
adisbladis, line 7946: These shouldn't be prefixed with `v`.

adisbladis (changes requested):

overrides/build-systems.json
Goorzhel: Whoops.

PR #1187, opened 2023-06-20
Update build-systems.json repath, flet, flet-core

jamesdbrock:
Is this the correct way repair these packages for problem #736 ?

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -5794,6 +5794,13 @@
   "fleep": [
     "setuptools"
   ],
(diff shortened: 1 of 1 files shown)


jamesdbrock:
Oh I see, the local workaround instructions are here. https://github.com/nix-community/poetry2nix/blob/master/docs/edgecases.md#modulenotfounderror-no-module-named-packagename

Anyway, this PR solves the problem upstream for these three packages.

jamesdbrock:
Ok rebased.

Issue #1139, opened 2023-04-26
pillow-9.5.0 fails to build


pmiddend:
### Describe the issue
I have `pillow = "9.5.0"` in my `pyproject.toml`, and it fails because:

```
python3.10-pillow>   The headers or library files could not be found for zlib,
python3.10-pillow>   a required dependency when compiling Pillow from source.
```

Adding `zlib` to buildInputs in my overrides as in:

```
pillow = super.pillow.overrideAttrs (old: { nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ self.zlib ]; });
```

Results in the same error.

I've checked the predefined overrides and saw that pillow is already included, see [here](https://github.com/nix-community/poetry2nix/blob/master/overrides/default.nix#L1669). 

With 9.4.0 (the previous version) poetry2nix worked fine, so something about their build process must have changed.

adisbladis:
This is because of an upstream issue: https://github.com/python-pillow/Pillow/pull/7069.

adisbladis:
We should apply that PR as a patch on 9.5.0.

asymmetric:
Or have it fixed in nixpkgs, see https://github.com/NixOS/nixpkgs/pull/228493

adisbladis:
> Or have it fixed in nixpkgs, see [NixOS/nixpkgs#228493](https://github.com/NixOS/nixpkgs/pull/228493)

This will not fix it in poetry2nix as we're not using nixpkgs definitions.

PR #1193, opened 2023-06-22
FAQ: Clarify poetry2nix does not use nixpkgs for Python packages

asymmetric:
I'm sure this could be expanded upon, and maybe this PR can provide the impetus for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,10 @@ in pkgs.poetry2nix.mkPoetryApplication {
 
 **A.** By default, poetry2nix installs from source. If you want to give precedence to wheels, look at the `preferWheel` and `preferWheels` attributes.
 
+**Q.** Does poetry2nix use package definitions from nixpkgs' Python package set?
+
+**A.** No.
+
 **Q.** How to prefer wheel installation for a single package?
 
 **A.** Override it and set `preferWheel = true` in that single package:
--- a/README.md
+++ b/README.md
@@ -317,7 +317,8 @@ in pkgs.poetry2nix.mkPoetryApplication {
 
 **Q.** Does poetry2nix use package definitions from nixpkgs' Python package set?
 
-**A.** No.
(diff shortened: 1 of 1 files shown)


takeda:
This is not strictly true. Many packages that aren't explicitly defined in the lock file are actually obtained from nixpkgs.

This mostly applies to overrides, things like when package needs setuptools, flit, poetry, hatch and of course non python dependencies like openssl etc.

asymmetric:
Interesting, I wrote this up after [a comment by @adisbladis](https://github.com/nix-community/poetry2nix/issues/1139#issuecomment-1602268969), but I might've interpreted it too broadly.

But in the end, this was my goal with this PR, to help spell out the architecture so we can document it somewhere visible.

takeda:
I see. The context matters. The question was about package (pillow) listed as a dependency to author's application. In that context pillow defined in poetry is used after applying overrides. The nixpkgs version is totally tossed away.

But then to install pillow poetry2nix adds setuptools dependency: https://github.com/nix-community/poetry2nix/blob/a98ecb470cc4b18407a156e17066173add8e05d3/overrides/build-systems.json#L11369-L11371

That dependency is not defined in poetry.lock file and is pulled from nixpkgs.

You might think "oh this is a special case and it only happens at build time", but then there are broken packages like ansible, where all its dependencies are pulled from nixpkgs because it doesn't define any:

https://github.com/nix-community/poetry2nix/blob/a98ecb470cc4b18407a156e17066173add8e05d3/overrides/default.nix#L111-L129

So the answer to this question is much more complex.

asymmetric:
So a better answer is that everything declared in a `poetry.lock` is built without relying on nixpkgs definitions, with the exception of under-specified packages, like e.g. ansible.

Is this accurate?

takeda:
The thing is that what is clear to me often is not clear to others, but this is what I would write:

> poetry2nix overlays packages taken from `poetry.lock` on top of nixpkgs (in such way that overwritten packages in nixpkgs are completely ignored). Any package that is used, that isn't in the `poetry.lock` file (most commonly build dependencies) is taken from nixpkgs.

PR #1194, opened 2023-06-26
open3d: don't include mesa on Darwin

bouk:
It's an optional dependency (AFAICT) and the current version of mesa is broken in nixpkgs


adisbladis:
Maybe we should guard this on `meta.broken` instead so we'll automatically pick up any fix from nixpkgs?

bouk:
@adisbladis thanks for picking it up :+1:, sorry

PR #1195, opened 2023-06-27
jsondiff override not needed anymore

takeda:
In version 2.0.0 released in April last year this override is not needed and causes issues when building from wheel.


adisbladis:
Could you add a version check? We'd still want the override for older versions.

PR #1201, opened 2023-07-05
overrides: discord-py

ProducerMatt:
This package doesn't build without setuptools.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -4029,6 +4029,9 @@
   "discogs-client": [
     "setuptools"
   ],
+  "discord-py": [
+    "setuptools"
+  ],
   "discordpy": [
     "setuptools"
   ],


adisbladis:
Could you add it via the JSON file as we do with other setuptools packages?

ProducerMatt:
@adisbladis done, sorry I missed that.

ProducerMatt:
I see there's already an override for `discordpy`, but I had to override `discord-py`, wonder why

ProducerMatt:
What's with the build failure? Is there a problem?

adisbladis:
> What's with the build failure? Is there a problem?

To reduce merge conflicts we require the build systems to be sorted.
This can be done with this one-liner:
`jq --raw-output --sort-keys < overrides/build-systems.json | sponge overrides/build-systems.json`

PR #1202, opened 2023-07-06
Fix override recommonmark and rich mixup

iwanb:
I was wondering why recommonmark did not appear in my environment :)

I've made that mistake a few times as well and it's rather confusing, is there a pattern which could avoid the name repetition?

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2308,7 +2308,7 @@ lib.composeManyExtensions [
         }
       );
 
-      recommonmark = super.rich.overridePythonAttrs (
+      recommonmark = super.recommonmark.overridePythonAttrs (
         old: {
           buildInputs = (old.buildInputs or [ ]) ++ [ self.commonmark ];
         }


adisbladis:
> I've made that mistake a few times as well and it's rather confusing, is there a pattern which could avoid the name repetition?

Nope. This sort of thing should be caught in code review but it's also very easy to overlook since these types of additions are so monotonic.

iwanb:
Thanks! Do you know if there's a way to workaround this without having to upgrade poetry2nix? I tried a few things like this but it doesn't seem to work:
```
          (self: super: (builtins.removeAttrs (poetry2nix.defaultPoetryOverrides self super) ["recommonmark"]))
```

adisbladis:
You can use `defaultPoetryOverrides.overrideOverlay (self: super: { recommonmark = super.recommonmark.overridePythonAttrs ( })`.
This mechanism takes precedence over the builtin overrides.

iwanb:
I tried that but it triggers infinite recursion somehow :/ Maybe because of the other overlay I have...

PR #1206, opened 2023-07-07
patch missing old buildinputs

jaybutera:
Safer overrides made fallbacks when buildInputs don't exist

https://github.com/nix-community/poetry2nix/pull/250/commits

But there were 3 missing cases I caught.


adisbladis:
Could you run nixpkgs-fmt?

PR #1209, opened 2023-07-07
Update overrides for `jsonschema` 4.18.0

sigprof:
[`jsonschema` 4.18.0](https://github.com/python-jsonschema/jsonschema/releases/tag/v4.18.0) gained a bunch of new dependencies which need overrides to build:
- [`rpds-py`](https://github.com/crate-py/rpds):
  - uses Rust and Cargo, and therefore the override would require updating the Cargo hash after every new release;
- [`referencing`](https://github.com/python-jsonschema/referencing):
  - uses the `hatchling` and `hatch-vcs` build system packages;
  - declares classifiers which are not accepted by `hatchling` 1.13.0 (the minimum requirements are apparently `hatchling` >= 1.14.0 and `trove-classifiers` >= 2023.4.25); updating `hatchling` would be really complicated, so the offending classifiers are patched out in `postPatch`;
- [`jsonschema-specifications`](https://github.com/python-jsonschema/jsonschema-specifications):
  - uses the `hatchling` and `hatch-vcs` build system packages;
  - also declares classifiers which are not accepted by `hatchling` 1.13.0 and need to be patched out.

Also the `jsonschema` 4.18.0 package itself declares the same problematic classifiers and therefore needs the same code in `postPatch`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -463,7 +463,6 @@ lib.makeScope pkgs.newScope (self: {
   /* Poetry2nix CLI used to supplement SHA-256 hashes for git dependencies  */
   cli = import ./cli.nix {
     inherit pkgs lib;
-    inherit (self) version;
   };
 
   # inherit mkPoetryEnv mkPoetryApplication mkPoetryPackages;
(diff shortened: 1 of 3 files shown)


sigprof:
The same commit as in #1210 is currently present here in order to pass the CI.

PR #1213, opened 2023-07-09
Add support for building Python 2 applications

williamvds:
By allowing the build-time Python runtime to be overridden.

This is handy for packaging legacy Python 2.7 applications with poetry2nix.

Convert the poetry2nix build-time inputs to `nativeBuildInputs` rather than `buildInputs`. This prevents mixed Python derivations from causing a "Python version mismatch in ..." error.

This is a succinct proof-of-concept to gauge interest, since it's not a huge change (in terms of line count) and I'd like to share this patch for anyone who's interested.

To do:

* [ ] optimise the `self.mkPoetryPackages` calls. The `buildPython` particularly needs the overlays which add `poetry` and `poetry-core`
  * find the minimal overlays for each runtime and refactor
* [ ] add a test for a Python 2 package
  * what would be a good candidate?
* [ ] support other `mkPoetry*` functions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Creates a Python application using the Python interpreter specified based on the
 - **overrides**: Python overrides to apply (_default_: `defaultPoetryOverrides`).
 - **meta**: application [meta](https://nixos.org/nixpkgs/manual/#chap-meta) data (_default:_ `{}`).
 - **python**: The Python interpreter to use (_default:_ `pkgs.python3`).
+- **buildPython**: The Python interpreter to use for building (_default:_ `python`).
 - **preferWheels** : Use wheels rather than sdist as much as possible (_default_: `false`).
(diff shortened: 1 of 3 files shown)


suntropez:
Just leaving a +1 that this would be nice to have.

williamvds:
Just adding a note that I've recently given up on using poetry2nix for Python 2, after a nixpkgs upgrade broke my package yet again.

Another gotcha I was dealing with was that recent Poetry versions cannot generate the lock files for Python 2 versions. I tried creating a package for the last version with Python 2 support, but didn't get anywhere with that.

cpcloud:
Should we close this out?

I don't really know what the policy is in poetry2nix for Python 2 support.

@adisbladis may know more

williamvds:
I don't intend to dedicate more time to this, apologies. I'll close this off, anyone who wishes to pick this up is welcome to do so

PR #1245, opened 2023-07-25
fix: also use `allRefs = true` when branch is given

cpcloud:
Fixes an issue where branches not created from upstream `HEAD` could not be found


cpcloud:
@adisbladis Thoughts here?

adisbladis:
> @adisbladis Thoughts here?

TBH not much. If it solves an issue for you I'm fine with it.

PR #1248, opened 2023-08-02
Update build-systems.json for sphinxcontrib-jquery

brendon-boldt:
First time contributor to this repo.  I am not sure of what the best method for testing this update is _in situ_.  I found that this override fixed an issue with building `sphinx-rtd-theme` with poetry2nix (i.e., `sphinxcontriib-jquery` not being able to find `flit_core`).

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -17138,6 +17138,9 @@
     "pbr",
     "setuptools"
   ],
+  "sphinx-js": [
+    "setuptools"
+  ],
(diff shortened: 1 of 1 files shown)


adisbladis:
This looks good except the build systems need to be sorted. You can use `jq --sort-keys`.

brendon-boldt:
Duh, I knew that and did it by hand but didn't go through the trouble of noting that "q" goes before "s". :facepalm:  Thanks for your patience...

PR #1257, opened 2023-08-14
overrides/pyqt5: be explicit about dependencies

K900:
Explicitly list the modules we depend on instead of relying on qt5.full. Also removes deprecated and unmaintained qt5.qtwebkit dependency.


goodboy:
Verified this works for installing `PyQt5` **without** webkit.

Will link PR once pushed 🏄🏼

goodboy:
Wait, does `preferWheel = true` support still work with this?

PR #1263, opened 2023-08-18
Fix various PyTorch-related libraries

MostAwesomeDude:
This is a cleaned-up pile of commits which lets me get most of the way towards building an application using PyTorch. Of note are the `safetensors` and `tokenizers` from HuggingFace, and `tiktoken` from OpenAI; these libraries have upstream issues, as noted in huggingface/tokenizers#1226 and openai/tiktoken#179.

This PR hardcodes two `Cargo.lock` files because upstreams are not willing to do it themselves. If that's not acceptable, then we'll need to figure out another permanent solution. However, I'm currently disinclined to accept any sort of demurral or delay, simply because the Nix way is to patch rather than to wait for upstreams.


MostAwesomeDude:
Huh. Apparently I broke `safetensors`. Clearly I tried to do too much in a single PR.

PR #1267, opened 2023-08-19
Respect `subdirectory` from `poetry.lock` when fetching sources

andersk:
Poetry 1.2 allows you to install from a subdirectory of a Git repository (python-poetry/poetry#5172). Fix poetry2nix to support this and add a test case.


tests/subdirectory/default.nix
```diff
@@ -0,0 +1,3 @@
+{ poetry2nix }:
+
+poetry2nix.mkPoetryEnv { projectDir = ./.; }
```
l0b0, line 3: Would an import test (see for example `tests/jq/default.nix`) be useful to strengthen the assertion this test is making?

l0b0 (approved):
Thanks!

tests/subdirectory/default.nix
andersk: The test fails at build time without subdirectory support, but sure, added a runtime import test.

l0b0 (approved):

Issue #1192, opened 2023-06-21
Can't install contourpy 1.1.0


andronat:
### Describe the issue

I'm getting this error when I'm trying to install `contourpy 1.1.0`:

```
Creating a wheel...
[33mWARNING: The directory '/homeless-shelter/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.[0m[33m
[0m[31mERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.[0m[31m
[0m
```

### Additional context

In my `flake.nix` I have the following, where `pkgs` point to `nixos-22.11` and  `newPkgs` to `nixos-23.05` as `meson 1.1.0` is required:
```nix
  contourpy = super.contourpy.overrideAttrs (old: {
    nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [
      newPkgs.meson
      pkgs.python310.pkgs.pybind11
      pkgs.python310.pkgs.meson-python
      pkgs.pkg-config
      pkgs.cmake
    ];
  });
```

`poetry.lock`:
```toml
[[package]]
name = "contourpy"
version = "1.1.0"
description = "Python library for calculating contours of 2D quadrilateral grids"
category = "main"
optional = false
python-versions = ">=3.8"
files = [
    {file = "contourpy-1.1.0-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:89f06eff3ce2f4b3eb24c1055a26981bffe4e7264acd86f15b97e40530b794bc"},
    {file = "contourpy-1.1.0-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:dffcc2ddec1782dd2f2ce1ef16f070861af4fb78c69862ce0aab801495dda6a3"},
    {file = "contourpy-1.1.0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:25ae46595e22f93592d39a7eac3d638cda552c3e1160255258b695f7b58e5655"},
    {file = "contourpy-1.1.0-cp310-cp310-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:17cfaf5ec9862bc93af1ec1f302457371c34e688fbd381f4035a06cd47324f48"},
[...]

PR #1268, opened 2023-08-21
overrides/contourpy: add dependency to meson on >=1.1.0

pineapplehunter:
I added `meson-python` and `pkg-config` as dependencies for contourpy when the version is >= 1.1.0.

The latest version of contourpy was not building with the older version of dependencies. It seems that they changed the build system. (see https://contourpy.readthedocs.io/en/latest/changelog.html#v1-1-0-2023-06-13)
This PR should fix the build issue.

Tested versions of contourpy:
- 0.0.5
- 1.0.7
- 1.1.0 (uses the newer build system)


pineapplehunter:
I just found this issue #1192.
This should fix it but, should I submit this to nixpkgs too?

PR #1281, opened 2023-08-23
allow usage of universal2 wheels

j-baker:
Have tested locally that this improves wheel selection.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -58,7 +58,9 @@ let
 
       rawDeps = pyProject.tool.poetry."dependencies" or { };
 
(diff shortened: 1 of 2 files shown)


cpcloud:
Do you have an example of a package that requires or uses this? If so, I can push up a test case for it to this PR.

cpcloud:
@j-baker Can you add a test case for this? I.e., build a package for which these changes are required?

Alternatively, if you can simply name a package you're working with that doesn't work without these changes I can add one for you.

adisbladis:
Superseded by #1387

PR #1284, opened 2023-08-23
pycurl: Add curl as a depedency

### LINKED ISSUE #1283: pycurl: add curl as a dependency
### Describe the issue
[pycurl](http://pycurl.io/)  fails to build because it requires `curl-config` at build time.
This means all source builds of pycurl and dependents fail to build with poetry2nix. This is meant as a tracking issues and I will open a PR with an override asap.

### Additional context
When using a dependent of pycurl, specifically oVirts Python SDK, pycurl is unable to complete the build stage.
`pyproject.toml`:
```toml
[tool.poetry]
name = "ovirt"
version = "0.1.0"
description = ""
authors = ["Joshua Kobschätzki <kobschaetzki@zib.de>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10"
ovirt-engine-sdk-python = "4.6.1"

[build-system]
requires = ["poetry-core"]
```
`nix log /nix/store/njxnmsgndms4s0smlvx1nmr7syzmbrgj-python3.11-pycurl-7.45.2.drv`:

```
Sourcing pip-install-hook
Using pipInstallPhase
Using pythonImportsCheckPhase
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
unpacking sources
unpacking source archive /nix/store/62xr7qi11fpkvxcj8wz6laq9b8p30cjp-pycurl-7.45.2.tar.gz
source root is pycurl-7.45.2
setting SOURCE_DATE_EPOCH to timestamp 1671219728 of file pycurl-7.45.2/setup.cfg
patching sources
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
no configure script, doing nothing
@nix { "action": "setPhase", "phase": "buildPhase" }
building
Executing pipBuildPhase
Creating a wheel...
[33mWARNING: The directory '/homeless-shelter/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.[0m[33m
[0mProcessing /build/pycurl-7.45.2
  Running command python setup.py egg_info
    File "/build/pycurl-7.45.2/setup.py", line 229, in configure_unix
      p = subprocess.Popen((self.curl_config(), '--version'),
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/nix/store/px3zbr30lbnmb3drlx5djwdd792l9vwz-python3-3.11.4/lib/python3.11/subprocess.py", line 1026, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/nix/store/px3zbr30lbnmb3drlx5djwdd792l9vwz-python3-3.11.4/lib/python3.11/subprocess.py", line 1950, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'curl-config'

  During handling of the above exception, another exception occurred:

    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "/build/pycurl-7.45.2/setup.py", line 970, in <module>
      ext = get_extension(sys.argv, split_extension_source=split_extension_source)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/build/pycurl-7.45.2/setup.py", line 634, in get_extension
      ext_config = ExtensionConfiguration(argv)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/build/pycurl-7.45.2/setup.py", line 93, in __init__
      self.configure()
    File "/build/pycurl-7.45.2/setup.py", line 234, in configure_unix
      raise ConfigurationError(msg)
  ConfigurationError: Could not run curl-config: [Errno 2] No such file or directory: 'curl-config'
  [1;31merror[0m: [1msubprocess-exited-with-error[0m
  
  [31m×[0m [32mpython setup.py egg_info[0m did not run successfully.
  [31m│[0m exit code: [1;36m1[0m
  [31m╰─>[0m See above for output.
  
  [1;35mnote[0m: This error originates from a subprocess, and is likely not a problem with pip.
  [1;35mfull command[0m: [34m/nix/store/px3zbr30lbnmb3drlx5djwdd792l9vwz-python3-3.11.4/bin/python3.11 -c '[0m
[34m  exec(compile('"'"''"'"''"'"'[0m
[34m  # This is <pip-setuptools-caller> -- a caller that pip uses to run setup.py[0m
[34m  #[0m
[34m  # - It imports setuptools before invoking setup.py, to enable projects that directly[0m
[34m  #   import from `distutils.core` to work with newer packaging standards.[0m
[34m  # - It provides a clear error message when setuptools is not installed.[0m
[34m  # - It sets `sys.argv[0]` to the underlying `setup.py`, when invoking `setup.py` so[0m
[34m  #   setuptools doesn'"'"'t think the script is `-c`. This avoids the following warning:[0m
[34m  #     manifest_maker: standard file '"'"'-c'"'"' not found".[0m
[34m  # - It generates a shim setup.py, for handling setup.cfg-only projects.[0m
[34m  import os, sys, tokenize[0m
[34m  [0m
[34m  try:[0m
[34m      import setuptools[0m
[34m  except ImportError as error:[0m
[34m      print([0m
[34m          "ERROR: Can not execute `setup.py` since setuptools is not available in "[0m
[34m          "the build environment.",[0m
[34m          file=sys.stderr,[0m
[34m      )[0m
[34m      sys.exit(1)[0m
[34m  [0m
[34m  __file__ = %r[0m
[34m  sys.argv[0] = __file__[0m
[34m  [0m
[34m  if os.path.exists(__file__):[0m
[34m      filename = __file__[0m
[34m      with tokenize.open(__file__) as f:[0m
[34m          setup_py_code = f.read()[0m
[34m  else:[0m
[34m      filename = "<auto-generated setuptools caller>"[0m
[34m      setup_py_code = "from setuptools import setup; setup()"[0m
[34m  [0m
[34m  exec(compile(setup_py_code, filename, "exec"))[0m
[34m  '"'"''"'"''"'"' % ('"'"'/build/pycurl-7.45.2/setup.py'"'"',), "<pip-setuptools-caller>", "exec"))' egg_info --egg-base /build/pip-pip-egg-info-xvjhqu7e[0m
  [1;35mcwd[0m: /build/pycurl-7.45.2/
  Preparing metadata (setup.py) ... [?25l[?25herror
[1;31merror[0m: [1mmetadata-generation-failed[0m

[31m×[0m Encountered error while generating package metadata.
[31m╰─>[0m See above for output.

[1;35mnote[0m: This is an issue with the package mentioned above, not pip.
[1;36mhint[0m: See above for details.
```

bzadm:
Adds support for pycurl via override. Tested on GNU/Linux with NixOS. Fixes #1283 

## Test

`pyproject.toml`:

```toml
[tool.poetry]
name = "pycurl-test"
version = "0.1.0"
description = ""
authors = ["nhr-zib <kobschaetzki@zib.de>"]

[tool.poetry.dependencies]
python = "^3.10"
pycurl = "7.45.2"

[build-system]
requires = ["poetry-core"]
```

`flake.nix`:
```nix
{
  description = "pycurl test";

  inputs = {
    flake-utils.url = "github:numtide/flake-utils";
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
    poetry2nix = {
      url = "github:bzadm/poetry2nix/master";
      inputs.nixpkgs.follows = "nixpkgs";
      inputs.flake-utils.follows = "flake-utils";
    };
  };

  outputs = { self, poetry2nix, nixpkgs, flake-utils }:
    flake-utils.lib.eachDefaultSystem (system:
      let pkgs = nixpkgs.legacyPackages.${system};
      in {
        devShells.default = pkgs.mkShell {
          buildInputs = [
            (poetry2nix.legacyPackages.${system}.mkPoetryEnv {
              pyproject = ./pyproject.toml;
              poetrylock = ./poetry.lock;
            })
          ];
        };
      });
}
```

`test.py`:
```python
# See http://pycurl.io/docs/latest/quickstart.html#retrieving-a-network-resource
import pycurl
import re
try:
    from io import BytesIO
except ImportError:
    from StringIO import StringIO as BytesIO

headers = {}
def header_function(header_line):
    # HTTP standard specifies that headers are encoded in iso-8859-1.
    # On Python 2, decoding step can be skipped.
    # On Python 3, decoding step is required.
    header_line = header_line.decode('iso-8859-1')

    # Header lines include the first status line (HTTP/1.x ...).
    # We are going to ignore all lines that don't have a colon in them.
    # This will botch headers that are split on multiple lines...
    if ':' not in header_line:
        return

    # Break the header line into header name and value.
    name, value = header_line.split(':', 1)

    # Remove whitespace that may be present.
    # Header lines include the trailing newline, and there may be whitespace
    # around the colon.
    name = name.strip()
    value = value.strip()

    # Header names are case insensitive.
    # Lowercase name here.
    name = name.lower()

    # Now we can actually record the header name and value.
    # Note: this only works when headers are not duplicated, see below.
    headers[name] = value

buffer = BytesIO()
c = pycurl.Curl()
c.setopt(c.URL, 'http://pycurl.io')
c.setopt(c.WRITEFUNCTION, buffer.write)
# Set our header function.
c.setopt(c.HEADERFUNCTION, header_function)
c.perform()
c.close()

# Figure out what encoding was sent with the response, if any.
# Check against lowercased header name.
encoding = None
if 'content-type' in headers:
    content_type = headers['content-type'].lower()
    match = re.search('charset=(\S+)', content_type)
    if match:
        encoding = match.group(1)
        print('Decoding using %s' % encoding)
if encoding is None:
    # Default encoding for HTML is iso-8859-1.
    # Other content types may have different default encoding,
    # or in case of binary data, may have no encoding at all.
    encoding = 'iso-8859-1'
    print('Assuming encoding is %s' % encoding)

body = buffer.getvalue()
# Decode using the encoding we figured out.
print(body.decode(encoding))
```

Test the build and functionality:

```bash
$ nix develop -i -c python test.py
[ ... expected HTML output ]
```

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2016,6 +2016,12 @@ lib.composeManyExtensions [
         }
       );
 
+      pycurl = super.pycurl.overridePythonAttrs (
+        old: {
+          propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
+        }
+      );
+
       pydantic-core = super.pydantic-core.override {
         preferWheel = true;
       };
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2018,7 +2018,7 @@ lib.composeManyExtensions [
 
       pycurl = super.pycurl.overridePythonAttrs (
         old: {
-          propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
+          buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.curl ];
           nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ pkgs.curl ];
         }
       );
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2019,6 +2019,7 @@ lib.composeManyExtensions [
       pycurl = super.pycurl.overridePythonAttrs (
         old: {
           propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
+          buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.curl ];
         }
       );
 
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2019,7 +2019,7 @@ lib.composeManyExtensions [
       pycurl = super.pycurl.overridePythonAttrs (
         old: {
           propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
-          buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.curl ];
+          nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ pkgs.curl ];
         }
       );
 


overrides/default.nix
```diff
@@ -1853,6 +1853,13 @@ lib.composeManyExtensions [
         }
       );
 
+      pycurl = super.pycurl.overridePythonAttrs (
+        old: {
+          propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
```
andersk, line 1858: This should not need to be propagated—after PycURL is built, the C headers for cURL are not needed to use PycURL.

overrides/default.nix
```diff
@@ -1853,6 +1853,13 @@ lib.composeManyExtensions [
...
+          nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ pkgs.curl ];
```
takeda, line 1859: Similarly to the above comment. I believe all you need is just specify `buildInputs` and add `pkgs.curl` there.

`nativeBuildInputs` is meant for tooling that are typically tools that are needed during build time. Its purpose is clearer when you do cross compiling. So you would specify there things like `cmake`, for python it would be `setuptools`, `poetry-core` etc.

andersk: This may need to be native for the `curl-config` script. (But still needn’t be propagated.)

adisbladis: Then the solution is to add curl as both a build input and a native input.

takeda: ah, my bad.

Chaostheorie: Thank you for the review and explanation regarding the build inputs. I'll try to make the changes tomorrow.

overrides/default.nix
```diff
@@ -1853,6 +1853,13 @@ lib.composeManyExtensions [
         }
       );
 
+      pycurl = super.pycurl.overridePythonAttrs (
+        old: {
+          buildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.curl ];
```
takeda, line 1858: @bzadm you should also rename `old.propagatedBuildInputs` to `old.buildInputs`

l0b0 (commented):
Thanks! Would you also be willing to create tests for this package? If so, see

 - `tests/default.nix` for the test entry and
 - the `tests/mutmut` directory for a module test or `tests/matplotlib-post-3-7` for a library test.

PR #1290, opened 2023-08-28
New bootstrap fixes

K900:
This gets things building, at least. We should probably figure out how to make these conditional on nixpkgs version though...

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1689068808,
-        "narHash": "sha256-6ixXo3wt24N/melDWjq70UuHQLxGV8jZvooRanIHXw0=",
+        "lastModified": 1692799911,
+        "narHash": "sha256-3eihraek4qL744EvQXsK1Ha6C3CR7nnT8X2qWap4RNk=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "919d646de7be200f3bf08cb76ae1f09402b6f9b4",
+        "rev": "f9e7cf818399d17d347f847525c5a5a8032e4e44",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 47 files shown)


overrides/default.nix
```diff
@@ -2760,14 +2760,6 @@ lib.composeManyExtensions [
         }
       );
 
-
-      # Stop infinite recursion by using bootstrapped pkg from nixpkgs
```
FRidh, line 2764: those using poerty2nix with python2 projects may want to keep this

tjni:
I would be happy to help fix what's broken too. I should have checked this before merging, but I forgot. With that said, I might need some hand holding since I don't use poetry2nix.

flake.lock
```diff
@@ -40,16 +40,16 @@
...
     "nixpkgs": {
       "locked": {
-        "lastModified": 1689192006,
-        "narHash": "sha256-QM0f0d8oPphOTYJebsHioR9+FzJcy1QNIzREyubB91U=",
+        "lastModified": 1693156971,
+        "narHash": "sha256-BE6BIXmVQyrOhkX4s/nP+/Xeipb1rRS1qLK/kpe7Pqk=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "2de8efefb6ce7f5e4e75bdf57376a96555986841",
+        "rev": "f900e733c1602b4c46064a9ef0514bb2345a8949",
```
lucc, line 47: I would say this makes it "conditional on nixpkg version". Otherwise there is `lib.version` which is a string :/

takeda:
What is the status of the fix? Do I need to include this patch if I want to use latest version of nixpkgs and poetry2nix?

delroth:
> What is the status of the fix? Do I need to include this patch if I want to use latest version of nixpkgs and poetry2nix?

You'll want to follow #1291 which has more info and status updates.

K900:
I really don't have the time or the spoons to invest into this anymore. Anyone interested in picking this up, please do.

K900:
I'll just close the PR because I really don't intend to work on this any time soon. Anyone willing to pick up the work is more than welcome to.

PR #1293, opened 2023-08-29
docs: add what-to-do for cargo/rust-related problems

Pegasust:


MostAwesomeDude:
I like this. Let's merge it.

I've encountered packages without `Cargo.lock`; what should we do for those? A holistic solution would be nice.

Pegasust:
I've never encountered rough edges for packages (call one `foo`) without `Cargo.lock` so I couldn't speak from experience. Most of the cases require maintainer's love to maintain a `Cargo.lock` for that specific package, either in a fork of `foo` or in `poetry2nix`.

You'd likely need to override the couple of Nix derivation stages for `foo-py` to incorporate this.

Sorry if it's too hand-wavy. Again, I haven't had success in patching those yet (gave up on https://github.com/openai/tiktoken and resorted on wheels)

PR #1303, opened 2023-09-05
chore: bump poetry version

cpcloud:
Bumps the version of poetry.


pkgs/poetry/update.nix
```diff
@@ -1,6 +1,5 @@
 let
-  sources = import ../../nix/sources.nix;
-  pkgs = import sources.nixpkgs {
+  pkgs = import <nixpkgs> {
```
cpcloud, line 2: @adisbladis Is there a better way to do this that doesn't depend on special path syntax?

cpcloud: I will rewrite this as a flake application in a separate PR and then rebase this PR on that one.

andronat:
epic!

brianmay (approved):

cpcloud:
Merging! If there are any issues we can always revert.

PR #1312, opened 2023-09-17
`httpcore`: `setuptools` -> `hatchling` in 0.18.0+

moonshxne:
Note that manually overriding the `buildInputs` in a flake actually results in:

```sh
last 10 log lines:
       > Executing pythonRemoveTestsDir
       > Finished executing pythonRemoveTestsDir
       > pythonCatchConflictsPhase
       > /nix/store/q2zpnf017gccs9w9mb0ikdqhkf0k5v82-catch_conflicts.py:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
       >   import pkg_resources
       > Found duplicated packages in closure for dependency 'hatchling':
       >   hatchling 1.18.0 (/nix/store/zjkajhjjz7wis9yvm3gwvphifx3hiav2-python3.10-hatchling-1.18.0/lib/python3.10/site-packages)
       >   hatchling 1.18.0 (/nix/store/2jy9c8hbk8v0rk45zv8680srs1jky977-python3.10-hatchling-1.18.0/lib/python3.10/site-packages)
       >
       > Package duplicates found in closure, see above. Usually this happens if two packages depend on different version of the same dependency.
```

so it's important that `build-systems.json` gets updated.

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -7228,7 +7228,18 @@
     "setuptools"
   ],
   "httpcore": [
-    "setuptools"
+    {
+      "buildSystem": "hatch-fancy-pypi-readme",
+      "from": "0.18.0"
+    },
+    {
+      "buildSystem": "hatchling",
+      "from": "0.18.0"
+    },
+    {
+      "buildSystem": "setuptools",
+      "until": "0.18.0"
+    }
   ],
   "httpie": [
     "setuptools"
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -7229,16 +7229,16 @@
   ],
   "httpcore": [
     {
+      "buildSystem": "setuptools",
+      "until": "0.18.0"
+    },
+    {
       "buildSystem": "hatch-fancy-pypi-readme",
       "from": "0.18.0"
     },
     {
       "buildSystem": "hatchling",
       "from": "0.18.0"
-    },
-    {
-      "buildSystem": "setuptools",
-      "until": "0.18.0"
     }
   ],
   "httpie": [


moonshxne:
Also, how are changes here reflected in the `poetry2nix` tool in `nixpkgs`? Before bumping into this issue, I've always used `pkgs.poetry2nix`, and I'd like to go back. In addition, I've bumped into segfaults when I make the poetry2nix here follow nixpkgs/nixos-unstable.

overrides/build-systems.json
```diff
@@ -7228,7 +7228,18 @@
...
+      "buildSystem": "hatch-fancy-pypi-readme",
+      "from": "0.18.0"
+    },
+    {
+      "buildSystem": "hatchling",
+      "from": "0.18.0"
+    },
+    {
+      "buildSystem": "setuptools",
+      "until": "0.18.0"
```
l0b0, line 7241: `until` entries should probably be before `from` entries, to keep the sequence is ascending order of versions. Other packages seem to have been done that way.

moonshxne: Thanks for your input! I actually copied the entry for `httpx` a couple of lines below, and it's ordered the same way (`from` entries before `until` entries), so maybe we could try to ensure consistency across the entire file?

l0b0: Consistency would be great. There is already a sorting step in CI; I've filed an [issue to strengthen it](https://github.com/nix-community/poetry2nix/issues/1320).

l0b0 (approved):

cpcloud (approved):

PR #1319, opened 2023-09-21
Additional build-systems entries

purcell:
For:

- `shandy-sqlfmt`
- `textual-dev`
- `textual-textarea`

(All of these are required to build https://github.com/tconbeer/harlequin, a TUI DuckDB IDE)

It seemed excessive to add all the test boilerplate for each individually, but I did it anyway since this has been requested in other recent PRs. Let me know if there was an easier route I should have followed.

The PR additionally changes the build-systems override for `py-tree-sitter`, which is not actually an installable python package. A nixpkgs entry with this name exists, but the pypi package is simply `tree-sitter`, and one of the above libraries depends on it.


purcell:
I'm still a committer here but it's been a while, so I'll request a cursory review rather than merging.

cpcloud:
Nice!

Are you using `harlequin` for something? 🙂

cpcloud:
Does `py-tree-sitter` need to remain? I assume it was added for some reason.

purcell:
> Does `py-tree-sitter` need to remain? I assume it was added for some reason.

I'm happy to leave it there, and perhaps that's safer, but what I *suspect* happened is that someone took all the `pythonPackages` entries in nixpkgs and added them to the overrides here. However, not all of those python packages in nixpkgs are actual installable PyPi packages — `py-tree-sitter`, for example, was what was built from the `py-tree-sitter` repo, and possibly pre-dated PyPi publishing of that code, but even that repo lists the installable package name as `tree_sitter`.

purcell:
I've just reverted the removal of `py-tree-sitter`, to be on the safe side.

purcell:
Also, hi again, nice to see you're doing fun data stuff in your new day job too.

cpcloud (approved):

purcell:
I was wondering what the deal is with syncing these overrides across to nixpkgs these days? Looks like a couple of months since it was last done. Can I help? Having the overrides arrive there would be a prerequisite for my [harlequin PR in nixpkgs](https://github.com/NixOS/nixpkgs/pull/256516).

Issue #1291, opened 2023-08-29
Segmentation fault


Defman:
> I'm not sure if this is the right place to post the issue or if it should be on the main nixos repository.

### Describe the issue

Running `nix build` or `nix flake check` causes a segmentation fault.

### Additional context
Tried on the following machines.
- OS X Arm64
- GitHub ubuntu runner

```shell
echo "experimental-features = nix-command flakes" >> /etc/nix/nix.conf
echo "sandbox = false" >> /etc/nix/nix.conf
echo "filter-syscalls = false" >> /etc/nix/nix.conf
```

`default.nix`/`shell.nix`/`flake.nix`:
```nix
{
  description = "Application packaged using poetry2nix";

  inputs.flake-utils.url = "github:numtide/flake-utils";
  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  inputs.poetry2nix = {
    url = "github:nix-community/poetry2nix";
    inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = { self, nixpkgs, flake-utils, poetry2nix }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        # see https://github.com/nix-community/poetry2nix/tree/master#api for more functions and examples.
        inherit (poetry2nix.legacyPackages.${system}) mkPoetryEnv defaultPoetryOverrides;
        pkgs = nixpkgs.legacyPackages.${system};
        pypkgs-build-requirements = {
          django-autoslug = [ "setuptools" ];
          django-hosts = [ "setuptools" ];
          django-autocomplete-light = [ "setuptools" ];
          django-bootstrap3 = [ "setuptools" ];
          django-classy-tags = [ "setuptools" ];
          crispy-bootstrap3 = [ "setuptools" ];
          django-easy-pdf3 = [ "setuptools" ];
          
        };
        p2n-overrides = defaultPoetryOverrides.extend (self: super:
          builtins.mapAttrs (package: build-requirements:
            (builtins.getAttr package super).overridePythonAttrs (old: {
              buildInputs = (old.buildInputs or [ ]) ++ (builtins.map (pkg: if builtins.isString pkg then builtins.getAttr pkg super else pkg) build-requirements);
            })
          ) pypkgs-build-requirements
        );
      in
      {
        packages = {
          myapp = mkPoetryEnv { 
            projectDir = self; 
            overrides = p2n-overrides;
            groups = [];
            checkGroups = [];
[...]

PR #1323, opened 2023-09-24
Nixpkgs backports

purcell:
There seem to have been some changes made to the poetry2nix overrides in nixpkgs over the last 3 months, and I guess we want them here too, in order to have this repo be the ultimate source of truth?

NixOS/nixpkgs@b9b3be8e099f
NixOS/nixpkgs@6b6390885543ee52d0569128648ffefb3d2439b7
NixOS/nixpkgs@253a291274c2790008fe57c97983d3581a2c7798

### DIFF
--- a/overrides/build-systems.json
+++ b/overrides/build-systems.json
@@ -1973,10 +1973,6 @@
     "setuptools",
     "setuptools-scm"
   ],
-  "backports-functools-lru-cache": [
-    "setuptools",
-    "setuptools-scm"
-  ],
   "backports-shutil-which": [
     "setuptools"
   ],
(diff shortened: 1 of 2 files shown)


purcell:
Alright, so the `wheel` change from downstream fails the tests here. Is this PR misguided? What's the plan for getting updates from here into nixpkgs now? It seems the release numbering scheme changed in June and there haven't been any syncs into nixpkgs from here since then.

andersk:
See also #1291; there was a corresponding update to `python3.pkgs.wheel`, so we need to update our nixpkgs commit at the same time.

purcell:
> so we need to update our nixpkgs commit at the same time.

Oh, I see, you mean the pinned version used in CI here. Makes sense. I'll add a commit here which bumps the flake deps, but I'll leave it to @adisbladis / @cpcloud to suggest how best to coordinate all this.

purcell:
Hmm, if that was indeed the suggestion, it didn't help.

cpcloud:
Closing this out in favor of #1329.

PR #1334, opened 2023-10-06
Add overrides for hidapi

### LINKED ISSUE #1019: Unable to build hidapi package
I've looked through the edgecases document and discussed this issue on the Poetry2nix matix chat.

I'm unable to get a project that depends on the hidapi package to build.
The simplest reproducible case:
pyproject.toml:
```toml
[tool.poetry]
name = "simple"
version = "0.1.0"
description = ""

[tool.poetry.dependencies]
python = "^3.10"
hidapi = "^0.13.1"


[build-system]
requires = ["poetry-core"]
```
---
flake.nix: (I used the template to make sure it was not my own bug)
```nix
{
  description = "Application packaged using poetry2nix";

  inputs.flake-utils.url = "github:numtide/flake-utils";
  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  inputs.poetry2nix = {
    url = "github:nix-community/poetry2nix";
    inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = { self, nixpkgs, flake-utils, poetry2nix }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        # see https://github.com/nix-community/poetry2nix/tree/master#api for more functions and examples.
        inherit (poetry2nix.legacyPackages.${system}) mkPoetryApplication;
        pkgs = nixpkgs.legacyPackages.${system};
      in
      {
        packages = {
          myapp = mkPoetryApplication { projectDir = self; };
          default = self.packages.${system}.myapp;
        };

        devShells.default = pkgs.mkShell {
          packages = [ poetry2nix.packages.${system}.poetry ];
        };
      });
}
```
---

nix build always fails with:
```
 gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -I/nix/store/c06np1spdb2bbsfp5x3716d529mrxw7b-libxcrypt-4.4.33/include -fPIC -Ihidapi/hidapi -I/usr/include/libusb-1.0 -I/nix/store/iw1vmh509hcbby8dbpsaanbri4zsq7dj-python3-3.10.10/incl>
  hidapi/libusb/hid.c:44:10: fatal error: libusb.h: No such file or directory
     44 | #include <libusb.h>
        |          ^~~~~~~~~~
  compilation terminated.
  error: command '/nix/store/bfbp3ypd9nm3fapz634gvvs738blrl0y-gcc-wrapper-12.2.0/bin/gcc' failed with exit code 1
  error: subprocess-exited-with-error

  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
```
---

I've tried tried various overrides (buildInputs, nativeBuildInputs) during some discussion on the matrix chat but nothing seems to fix this.
One example in case I'm doing something obviously wrong:
```nix
          myapp = mkPoetryApplication {
            projectDir = self;
            overrides = defaultPoetryOverrides.extend
            ( self: super: {
              hidapi = super.hidapi.overridePythonAttrs
              (
                old: {
                  buildInputs = (old.buildInputs or [ ]) ++ [
                    pkgs.libusb1
                    pkgs.udev
                  ];
                }
              );
            });
          };
```

PicoGeyer:
resolves #1019

Patch found in nixpkgs for hidapi

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -857,7 +857,7 @@ lib.composeManyExtensions [
 
       hidapi = super.hidapi.overridePythonAttrs (
         old: {
-          buildInputs = old.nativeBuildInputs or [ ] ++ [
+          propagatedBuildInputs = old.propagatedBuildInputs or [ ] ++ [
             pkgs.libusb1
           ];
           postPatch = lib.optionalString stdenv.isLinux ''
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -855,6 +855,19 @@ lib.composeManyExtensions [
         }
       );
 
+      hidapi = super.hidapi.overridePythonAttrs (
+        old: {
+          buildInputs = old.nativeBuildInputs or [ ] ++ [
+            pkgs.libusb1
+          ];
+          postPatch = lib.optionalString stdenv.isLinux ''
+            libusb=${pkgs.libusb1.dev}/include/libusb-1.0
+            test -d $libusb || { echo "ERROR: $libusb doesn't exist, please update/fix this build expression."; exit 1; }
+            sed -i -e "s|/usr/include/libusb-1.0|$libusb|" setup.py
+          '';
+        }
+      );
+
       hikari = super.hikari.overrideAttrs (
         old: {
           buildInputs = (old.buildInputs or [ ]) ++ [ self.setuptools ];


overrides/default.nix
```diff
@@ -855,6 +855,19 @@ lib.composeManyExtensions [
         }
       );
 
+      hidapi = super.hidapi.overridePythonAttrs (
+        old: {
+          buildInputs = old.nativeBuildInputs or [ ] ++ [
```
prusnak, line 860: Shouldn't this be `nativeBuildInputs`?

```suggestion
          nativeBuildInputs = old.nativeBuildInputs or [ ] ++ [
```

PicoGeyer: Sorry for that mistake, but no, I think it should be buildInputs like so
```suggestion
          buildInputs = old.buildInputs or [ ] ++ [
```
Otherwise there is a linking error, libusb1 will be used at run time. 
I see that nixpkgs actually has it as propagatedBuildInputs, Is it better to use that?

prusnak: Right. I think it should belong to `propagatedBuildInputs` afterall.

overrides/default.nix
```diff
@@ -855,6 +855,19 @@ lib.composeManyExtensions [
...
       );
 
+      hidapi = super.hidapi.overridePythonAttrs (
+        old: {
+          propagatedBuildInputs = old.propagatedBuildInputs or [ ] ++ [
+            pkgs.libusb1
+          ];
+          postPatch = lib.optionalString stdenv.isLinux ''
+            libusb=${pkgs.libusb1.dev}/include/libusb-1.0
+            test -d $libusb || { echo "ERROR: $libusb doesn't exist, please update/fix this build expression."; exit 1; }
```
cpcloud, line 865: Is this test necessary?

PicoGeyer: I copied this from nixpkgs:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hidapi/default.nix

I'm guessing it just makes it a bit more clear where the failure is, if the include/libusb-1.0 directory is no longer there (perhaps due to a version bump)

PR #1341, opened 2023-10-10
blinker uses flit-core from 1.6.3

ksonj:


l0b0 (changes requested):
It's not strictly necessary, but would you be willing to create a test for this package? If so, see

- `tests/default.nix` for the test entry and
- the `tests/mutmut` directory for a module test or `tests/matplotlib-post-3-7` for a library test.

A reference, such as a link to the relevant change log entry, in the commit message would also be useful.

cpcloud:
Closing in favor of #1344

PR #1344, opened 2023-10-11
Configure build system for blinker >= 1.6.3

apeschar:
Same as #1341, but with tests, and the exclusive `until` set properly.

I didn't notice the other PR until I nearly finished this.


overrides/build-systems.json
```diff
@@ -2285,7 +2285,14 @@
...
   ],
   "blinker": [
-    "setuptools"
+    {
+      "buildSystem": "setuptools",
+      "until": "1.6.3"
+    },
+    {
+      "buildSystem": "flit-core",
+      "from": "1.6.2"
```
l0b0, line 2294: This looks like the opposite error: `from` is *inclusive*, so this should be `"1.6.3"`. By the way, it's generally better to just update a PR (even rewriting history) than creating a new PR with the same purpose.

l0b0: Oh whoops, I assumed you were the same author!

l0b0 (approved):
Very nice, thank you!

cpcloud (approved):

PR #1348, opened 2023-10-17
Add dependency on native Tesseract OCR executable for pytesseract

TikhonJelvis:
The pytesseract package needs to have the `tesseract` executable available at runtime to work.

By default, the Python package looks for the `tesseract` executable in the PATH. This doesn't work here, so we need to override the `tesseract_cmd` with the path to the `tesseract` executable we pulled in from Nix. I did this with a patch [based on how pytesseract is set up in Nixpkgs][1].

The patching code feels a bit fiddly. I don't know the idiomatic way to do this sort of thing.

I included a test that will fail if pytesseract cannot find the `tesseract` executable. The test passed for me with both `preferWheels = true` and `preferWheels = false`, but I only included one in the test suite here, not sure if it makes sense to have both—the actual patching code had to be a bit different depending on whether the source was a wheel or not.

[1]: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/pytesseract/tesseract-binary.patch


overrides/default.nix
```diff
@@ -2141,6 +2141,34 @@ lib.composeManyExtensions [
...
+
+            -tesseract_cmd = 'tesseract'
+            +tesseract_cmd = '${pkgs.tesseract4}/bin/tesseract'
+
+             numpy_installed = find_loader('numpy') is not None
+             if numpy_installed:
+          '';
+        in
+        super.pytesseract.overridePythonAttrs (old: {
+          buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.tesseract4 ];
```
l0b0, line 2161: Does this not need to be a `propagatedBuildInput`, to be available at runtime? In that case you might not need the patch.

TikhonJelvis: Good question, let me try it. I saw that the [pytesseract default.nix in Nixpkgs][1] used `buildInputs`, but I don't know if that's the best way to do it.

For reproducibility it seems better to have an explicit path to the executable rather than relying on the `PATH` variable where the script runs, so maybe that's the consideration.

[1]: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/pytesseract/default.nix

TikhonJelvis: Okay, I tried changing it to `propagatedBuildInputs`  and removing the patch, and the test case I wrote failed:

```
pytesseract.pytesseract.TesseractNotFoundError: tesseract is not installed or it's not in your PATH. See README file for more information.
```

cpcloud (approved):

PR #1371, opened 2023-10-26
cleanup getcargohash

cpcloud:
This PR attempts to remove `getCargoHash` from overrides and use the fetched source Cargo.lock file where possible.


adisbladis:
This is IFD. We can't do this.

adisbladis:
I would love to clean up cargo hashes in general. Maybe through a CI bot?

PR #1384, opened 2023-10-28
chore: ensure that ci fails when matrix construction errors

cpcloud:
Fixes CI matrix construction and overlay handling.


cpcloud:
Will add the overlay fix once I see that the matrix construction is failing with `master`

PR #1387, opened 2023-10-30
mk-poetry-dep: Use wheel platform filtering from pyproject.nix 

adisbladis:

Closes #1281


adisbladis:
Wheel filtering works fine but we also need to sort the list before taking the first element.

adisbladis:
Note that the Darwin failure isn't actually a new CI failure. 
It's caused by a previous correctness issue where SDK versions were not taken into account when checking for wheel compatibility with the Darwin platform.

`x86_64-darwin` is stuck on SDK 10.12 which has very poor compatibility with wheels in the Python ecosystem.

PR #1431, opened 2023-11-25
pip-build-hook: enable validation of build dependencies

Kiskae:
Per https://pip.pypa.io/en/stable/cli/pip_wheel/#cmdoption-check-build-dependencies

It turns out that unlike pypa, pip only checks build dependencies if it is told to.

This makes the missing build dependencies a lot clearer, since it immediately lists all missing dependencies rather than fail when it can't find them.

----

This change turns out to have surfaced a few hidden bugs with existing test cases, additional commits fix these issues.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1245,7 +1245,8 @@ lib.composeManyExtensions [
           # jupyterlab?)
           postPatch = ''
             substituteInPlace pyproject.toml \
-              --replace ', "jupyterlab~=3.0"' ""
+              --replace ', "jupyterlab~=3.0"' "" \
+              --replace ', "jupyterlab~=4.0"' ""
           '';
         }
       );
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1886,7 +1886,12 @@ lib.composeManyExtensions [
 
         # For OSX, we need to add a dependency on libcxx, which provides
         # `complex.h` and other libraries that pandas depends on to build.
-        postPatch = lib.optionalString (!(old.src.isWheel or false) && stdenv.isDarwin) ''
+        postPatch = ''
+          if [ -f pyproject.toml ]; then
+            substituteInPlace pyproject.toml \
+              --replace 'meson-python==0.13.1' 'meson-python'
+          fi
+        '' + lib.optionalString (!(old.src.isWheel or false) && stdenv.isDarwin) ''
           if [ -f setup.py ]; then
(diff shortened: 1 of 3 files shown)


l0b0 (approved):
Thank you! Would you be OK to include a copy of a relevant error message in the ticket, for reference?

Kiskae:
> Thank you! Would you be OK to include a copy of a relevant error message in the ticket, for reference?

Added an example to the issue body, getting it to trigger intentionally was trickier than I expected

overrides/default.nix
```diff
@@ -1886,7 +1886,12 @@ lib.composeManyExtensions [
 
         # For OSX, we need to add a dependency on libcxx, which provides
         # `complex.h` and other libraries that pandas depends on to build.
-        postPatch = lib.optionalString (!(old.src.isWheel or false) && stdenv.isDarwin) ''
+        postPatch = ''
+          if [ -f pyproject.toml ]; then
```
l0b0, line 1890: Rather than checking for the presence of a file in Bash I think it's preferable to check for the package version in the Nix code. Doing it that way means we can remove the check once the older version is definitely no longer relevant.

Kiskae: Its been a while since I've looked at this PR, afaik I did it this way to mimic the existing `setup.py` patch.

Regarding versions the problematic dependency is still the same on the primary development branch, so it needs to be applied to for all versions:
https://github.com/pandas-dev/pandas/blob/4921e0b2290f881e17bd21685289e46ae4894155/pyproject.toml#L5

cpcloud:
One difficulty here is that this check enforces version constraints, which I think might be a non-starter without some `substituteInPlace` calls in the right place.

cpcloud:
Another issue is that this flag results in a failure for packages that are missing build time dependencies that are circular. In some cases we have to remove those build time dependencies to prevent infinite recursion.

cpcloud:
I do think there's value in checking for **missing** build-time dependencies but I'm not sure how to surface only those and ignore the errors coming from constraint violations.

overrides/default.nix
cpcloud: This seems fine for now.

cpcloud:
Happy to accept the changes other than the build hook modification, merge those shortly!

cpcloud (approved):

cpcloud:
Passing locally on:

- x86_64-darwin
- aarch64-darwin
- x86_64-linux

PR #1481, opened 2024-01-02
fix(pytorch-lightning): disable unpackPhase override

cameronraysmith:
There are default overrides for  `pytorch-lightning`  that attempt to handle what must have previously been something like a nested tarball. I removed these and what previously resulted in

```bash
unpacking sources
tar: This does not look like a tar archive
```

during installation proceeded as expected. This is at least true when I source the boundaries of the `[1.9.4 2.1.3]` version range in a test flake on Darwin aarch64 and Linux x86_64 systems. I am posting this PR to mark the issue and discuss the necessity of retaining these `unpackPhase` customizations with version guards for backward compatibility versus removing them altogether.

I could add such guards if I knew where things changed, but I didn't manage to bisect where that is in the pytorch-lightning release history after experimenting with `1.9.4` and sporadically downloading a few of the very early source distributions from PyPI to inspect their structure.

Since the `unpackPhase` was added in [poetry2nix #786 Oct 31 2022](https://github.com/nix-community/poetry2nix/pull/786) the change should presumably be somewhere between the `pytorch-lightning` release prior to that [Sep 22 2022 1.7.7](https://pypi.org/project/pytorch-lightning/1.7.7/#files) and the oldest one I checked manually from [March 2 2023 in 1.9.4](https://pypi.org/project/pytorch-lightning/1.9.4/#files).

Alternatively, there may have been a higher-level change in how the source tarball's are handled rendering this unpackPhase unnecessary, even for older versions.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -2432,22 +2432,11 @@ lib.composeManyExtensions [
         '';
       });
 
-
       pytoml = super.pytoml.overridePythonAttrs (
         _old: {
           doCheck = false;
         }
       );
-
-      pytorch-lightning = super.pytorch-lightning.override {
-        unpackPhase = ''
-          # $src remains a gzipped tarball otherwise.
-          mkdir -p tmp
-          tar xvf $src --directory=tmp
-          mv tmp/pytorch-lightning*/* .
-          rm -rf tmp
-        '';
-      };
 
       pyqt5 =
         let


knkski:
I checked every major x.y.0 release such as 1.4.0, 1.5.0, all the way back to the first release with a wheel, which was 0.7.3. I also exhaustively checked 1.7.7 to 1.9.4. I was unable to find any versions that had this nested tarball behavior.

I don't know how pypi handles things exactly, but if I had to guess, I'd say that the problem was fixed on their end somehow.

cameronraysmith:
I might be totally incorrect in having proposed a nested tarball might have been the reason for this.

knkski:
Sorry, to clarify, I mean that I tried to install all of those versions with the latest commit of `poetry2nix`, and they all failed with the same `tar: This does not look like a tar archive`. I was unable to find a version that the unpackPhase override would seem to apply to.

PR #1547, opened 2024-03-02
overrides: Fix marshmallow and twisted

ostrolucky:
I've experienced problems with building https://github.com/sp1thas/criticker-dataset without this


cpcloud:
These are both fixed in `master` now.

PR #1550, opened 2024-03-05
nvidia: use different method for making libraries work

bouk:
This symlinks all the .so libraries to $out/lib/ so the autopatchelf hook can find them, instead of relying on torch's inbuilt magic to find the libraries.

I've also formatted it as a separate overlay that generates the override for all the nvidia packages.

### DIFF
--- a/overrides/default.nix
+++ b/overrides/default.nix
@@ -1792,39 +1792,27 @@ lib.composeManyExtensions [
               ];
             }) else prev.notebook;
 
-      nvidia-cudnn-cu11 = super.nvidia-cudnn-cu11.overridePythonAttrs (attrs: {
+      nvidia-cudnn-cu11 = prev.nvidia-cudnn-cu11.overridePythonAttrs (attrs: {
         propagatedBuildInputs = attrs.propagatedBuildInputs or [ ] ++ [
-          self.nvidia-cublas-cu11
+          final.nvidia-cublas-cu11
         ];
       });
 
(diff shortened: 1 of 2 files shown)


overrides/default.nix
```diff
@@ -3755,4 +3722,42 @@ lib.composeManyExtensions [
...
+    "nvidia-nvjitlink-cu11"
+    "nvidia-nvjitlink-cu12"
+    "nvidia-nvtx-cu11"
+    "nvidia-nvtx-cu12"
+  ] (name: super.${name}.overridePythonAttrs (old: {
+    # 1. Remove __init__.py because all the cuda packages include it
+    # 2. Symlink the cuda libraries to the lib directory so autopatchelf can find them
+    postFixup = ''
+      rm -r $out/${self.python.sitePackages}/nvidia/{__pycache__,__init__.py}
+      ln -sfn $out/${self.python.sitePackages}/nvidia/*/lib/lib*.so* $out/lib
```
bouk, line 3759: Maybe this is the wrong approach: there's also a `addAutoPatchelfSearchPath` function that can be used to just add libs for that specific package. But I don't know how to make that work for all of its dependencies without listing them all explicitly.

This approach works at least and it's straightforward.

cpcloud:
@bouk Any chance you can add a test for this? I'm guessing that a simple test package that depends on torch >= 2.0.0 would suffice.

PR #1551, opened 2024-03-05
overrides: PyQt6 v6.6.1 ignore libmimerapi.so

jamesdbrock:
I needed this to upgrade PyQt6 from v6.5.3 to v6.6.1.


jamesdbrock:
I'm puzzled about why the checks are failing and I'd like advice please.

considerate:
I don't know which is the offending package but it looks like this is the actual error:

```
error: sphinx-7.2.6 not supported for interpreter python3.8
```

PR #1560, opened 2024-03-09
Fix infinite recursion if .git directory doesn't exist

TSRBerry:
This cost me a few hours yesterday when I was trying to figure out why my flake wasn't working anymore.
Turns out all the files and directories of my gitignore (including the `.git` directory) were already filtered out by the time `findGitIgnores` runs.

This shouldn't be a problem as long as it eventually aborts, but that wasn't the case.
Since each iteration adds another `/..` to the path it can't ever be equal to `/` which causes the function to recurse forever.

By using `dirOf` instead of appending `/..` we can avoid this issue and make sure this function always aborts at some point.

I'd like to add some tests for this, but I have no idea how to do that as I'm still new to Nix. Could you help me with that?

---


cpcloud:
@TSRBerry I can help add a test if you can describe a directory structure that would trigger the recrusion.

TSRBerry:
For the directory structure it's only important that none of the parent directories contain a .gitignore or .git directory

cpcloud:
@TSRBerry Thanks, I've got a failing-with-`master` test case now, that passes with this change.

cpcloud:
Passing locally for me on

- x86_64-linux
- x86_64-darwin
- aarch64-darwin

TSRBerry:
Awesome, thank you!

PR #1562, opened 2024-03-11
overrides: add types-oauthlib and types-xmltodict

JaniM:


cpcloud:
I'll push up another PR since you don't allow maintainers to edit forks.

PR #1574, opened 2024-03-25
feat: Split flake template devShell: app and poetry

jamesdbrock:
The nix template generates two devShells instead of one.

- The first devShell provides project dependencies, but not poetry.
- The second devShell provides only poetry.

In all of my **poetry2nix** projects this is the pattern that I use. I think this should be the default template `flake.nix`.

### DIFF
--- a/templates/app/flake.nix
+++ b/templates/app/flake.nix
@@ -23,8 +23,21 @@
           default = self.packages.${system}.myapp;
         };
 
+        # Shell for app dependencies.
+        #
+        #     nix develop
+        #
+        # Use this shell for developing your app.
         devShells.default = pkgs.mkShell {
           inputsFrom = [ self.packages.${system}.myapp ];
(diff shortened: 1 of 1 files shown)


cpcloud:
Any particular reason you prefer this setup as opposed to a devShell that contains environment dependencies as well as poetry? Personally, I like to have everything in a single mega environment.

jamesdbrock:
By the way thank you @cpcloud for finishing #1559 :heart:

cpcloud:
> I suppose I could do nix run nixpkgs#poetry but then I get a random version of poetry.

Yeah. Good point. I end up usually creating some scripts for this, but having a separate shell might be cleaner.

I suppose an alternative would be to have a separate `poetry` app in the flake whose sole job is to run poetry commands.

cpcloud (approved):

Issue #1553, opened 2024-03-06
stuck attempting to package `vllm 0.3.3`


meditans:
Hi! I'm attempting to package the latest version of `vllm` using poetry2nix.

I believe I'm mostly there: I've patched numerous dependencies that either lacked build tools or, like `safetensors`, relied on `maturin`. Thus, I used `preferWheel`. Currently, I'm addressing the last package, `vllm` itself, but I'm encountering a few unresolved errors.

Please find the work I've done thus far in this autocontained flake: [packaging-vllm](https://github.com/meditans/packaging-vllm).
I would appreciate ideas on how to finish packaging this, or directions on how to approach these problems. Then I could start upstreaming the patches that are needed to poetry2nix.

- In the master branch, I've tried to minimize the reliance on wheels. Here, I'm experiencing an issue with the `CUDA_HOME` variable possibly not being correctly set, and an incorrect version of `stdenv` (which I'm attempting to fix).
  
- In the `only-wheels` branch, I've focused on exclusively using wheels. However, I’ve encountered collision issues that I'm unsure how to resolve.

The errors are immediately visible if issuing `nix develop .#` in the respective branch. I'm on NixOS and I have some cuda GPUs.
Thank you :)

PR #1575, opened 2024-03-25
overrides: add fixups for vllm's both wheel and source mode

Atry:


Atry:
This PR partially fixes #1553

overrides/default.nix
```diff
@@ -1366,7 +1378,7 @@ lib.composeManyExtensions [
           );
           llvm = pkgs."llvmPackages_${llvm_version}".llvm or (throw "LLVM${llvm_version} has been removed from nixpkgs; upgrade llvmlite or use older nixpkgs");
         in
-        {
+        lib.optionalAttrs (!old.src.isWheel or false) {
```
cpcloud, line 1381:
```suggestion
        lib.optionalAttrs (!(old.src.isWheel or false)) {
```

overrides/default.nix
```diff
@@ -1664,6 +1676,12 @@ lib.composeManyExtensions [
         }
       );
 
+      numba = super.numba.overridePythonAttrs (
+        old: {
+          propagatedBuildInputs = (old.propagatedBuildInputs or [ ]) ++ [ pkgs.tbb_2021_8 ];
```
cpcloud, line 1681:
```suggestion
          propagatedBuildInputs = old.propagatedBuildInputs or [ ] ++ [ pkgs.tbb_2021_8 ];
```

cpcloud: Did tbb become a hard requirement for numba at some point?

overrides/default.nix
```diff
@@ -3782,6 +3836,31 @@ lib.composeManyExtensions [
       pydantic = super.pydantic.overridePythonAttrs
         (old: { buildInputs = old.buildInputs or [ ] ++ [ pkgs.libxcrypt ]; });
 
+      vllm = super.vllm.overridePythonAttrs (old: {
+        autoPatchelfIgnoreMissingDeps = true;
+      } // (lib.optionalAttrs (!old.src.isWheel or false) rec {
```
cpcloud, line 3841:
```suggestion
      } // (lib.optionalAttrs (!(old.src.isWheel or false)) rec {
```

Atry:
@cpcloud The test failure seems not due to this PR

overrides/default.nix
```diff
@@ -1705,6 +1717,12 @@ lib.composeManyExtensions [
         }
       );
 
+      numba = prev.numba.overridePythonAttrs (
+        old: {
+          propagatedBuildInputs = old.propagatedBuildInputs or [ ] ++ [ pkgs.tbb_2021_11 ];
```
cpcloud, line 1722: Is the TBB dependency necessary?

Atry: Yes

cpcloud: I don't think TBB is required: https://numba.readthedocs.io/en/stable/user/threading-layer.html#which-threading-layers-are-available

There are multiple ways to back the threading layer.

If you want to TBB with numba with poetry2nix, it's up to the end user to add tbb as a dependency. We shouldn't opt people in to optional dependencies like this. Please revert the change.

cpcloud: It seems like the requirement for tbb is coming from vllm, so numba should be made to ignore missing autopatchelf deps and then you'll have to make sure in your application that uses vllm that you depend on tbb in some way.

cpcloud: Fixed.

cpcloud (approved):

Atry (commented):
Looks good to me

Atry:
vllm 0.3.3 works, but 0.4.1 crashes https://github.com/nix-community/poetry2nix/pull/1635

PR #1618, opened 2024-05-01
Add `subdirectory` test for `git` dependency

cpcloud:
Redux of #1584 rebased on `master`.


cpcloud:
@jvanvugt The issue you're observing in #1584 isn't related to poetry2nix. The error you're getting is from poetry itself, due to an invalid specification.
j
`subdirectory` can only be used along with `git`. You can't specify `subdirectory` **without also specifying `git`**. I'll add a test case in this PR, but I'm going to revert your change to remove the `subdirectory` field unless you have a correctly failing example.

cpcloud:
Apparently you also **cannot** specify the `version` field either.

PR #1621, opened 2024-05-01
fix(pyside6): add test for pyside6

cpcloud:
Supercede #1494. This test passes for me on both x86_64-darwin and aarch64-darwin locally.


cpcloud:
@CyborgPotato Can you try this out? I wasn't able to get your original PR to build, but it seems like the test you added in #1494 passes when running against `master`.

cpcloud:
Later versions of pyside work on aarch64-darwin, but not x86_64-darwin 🤷🏻 

Keeping it at the older version in the test suite for now.

PR #1625, opened 2024-05-02
ci: allow more concurrency in CI

cpcloud:
Allow builds that do not require one-at-a-time runs to run concurrently

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: generate poetry.lock
         working-directory: ${{ env.TESTDIR }}
-        run: nix develop -c poetry lock --no-update
+        run: nix develop '.#poetry' -c poetry lock --no-update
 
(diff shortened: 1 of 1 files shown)


zowoq:
Concurrency was intentionally disabled in https://github.com/nix-community/poetry2nix/pull/1470 because this repo runs 150 jobs at once which blocks jobs in other repos.

cpcloud:
Oh, because of the shared resources in all of `nix-community`? Ok, I can disable it again, sorry about that.

cpcloud:
FWIW, 150+ jobs are still run with or without this PR, the difference is that individual jobs can be queued from other PRs, instead of only allowing a single workflow to run at once. Allowing only a single workflow's worth of jobs to run definitely cuts down on the amount of concurrency used by this repo though.

zowoq:
https://nix-community.org/continuous-integration/

Using the community buildbot might be an option for the builds that don't need overridden inputs, etc.

PR #1635, opened 2024-05-06
fix(vllm): fix up vllm 0.4.2

Atry:


Atry:
https://github.com/NixOS/nixpkgs/pull/309477 is needed to investigate the circular dependencies between fastapi and fastapi-cli.

cpcloud:
@Atry I added an override for `vllm-nccl-cu12` that downloads the rather large SO file and allows building vllm without the user having to hack in an environment variable, or download things from the internet in a random setup.py script :sweat_sm

Atry:
Copying the so file to a path under tmpdir does not make sense. It would get discarded immediately.

overrides/default.nix
```diff
@@ -3556,8 +3556,7 @@ lib.composeManyExtensions [
       # Circular dependency between triton and torch (see https://github.com/openai/triton/issues/1374)
       # You can remove this once triton publishes a new stable build and torch takes it.
       triton = prev.triton.overridePythonAttrs (old: {
-        propagatedBuildInputs = (builtins.filter (e: e.pname != "torch") old.propagatedBuildInputs) ++ [
-
+        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) [ final.torch ] ++ [
```
Atry, line 3559:
```suggestion
        propagatedBuildInputs = (removePackagesByName (old.propagatedBuildInputs or [ ]) [ final.torch ]) ++ [
```

Atry (commented):
`triton` needs `setuptools`

overrides/default.nix
```diff
@@ -881,11 +881,11 @@ lib.composeManyExtensions [
 
       fastapi = prev.fastapi.overridePythonAttrs (old: {
         # fastapi 0.111 depends on fastapi-cli, which depends on fastapi, resulting in infinite recursion
-        propagatedBuildInputs = builtins.filter (e: e.pname != "fastapi-cli") old.propagatedBuildInputs;
+        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) [ final.fastapi-cli ];
```
Atry, line 884: `final.fastapi-cli` exists only for `fastapi` version 0.111.0. The previous versions of `fastapi` do not depend on `final.fastapi-cli`.

overrides/default.nix
```diff
@@ -3981,13 +3980,30 @@ lib.composeManyExtensions [
...
+          soVersion = "2.18.1";
+          soName = "libnccl.so.${soVersion}";
+          libncclSO = pkgs.fetchurl {
+            url = "https://github.com/vllm-project/vllm-nccl/releases/download/v0.1.0/cu12-${soName}";
+            hash = "sha256-AFUDtiuf5Ga2svYLoq0dqkVxGW/8uCo9d6PHdb/NWsg=";
+          };
+        in
+        {
+          preBuild = ''
+            export HOME="$(mktemp -d)"
```
Atry, line 3998: To include the SO file, we need
``` suggestion
export HOME="$out"
```

In addition, `vllm` should be patched to locate the SO.

overrides/default.nix
Atry:
```suggestion
        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) lib.optional final?fastapi-cli final.fastapi-cli;
```

overrides/default.nix
```diff
@@ -888,8 +888,13 @@ lib.composeManyExtensions [
         }
       );
 
+      fastapi = prev.fastapi.overridePythonAttrs (old: {
+        # fastapi 0.111 depends on fastapi-cli, which depends on fastapi, resulting in infinite recursion
```
cpcloud, line 892: Why are you using `final ? fastapi-cli`? You should always use `final.fastapi-cli`.

overrides/default.nix
cpcloud: What would this change accomplish?

cpcloud: I think I see.

The library will end up in a place that would allow `vllm` to be patched to use it.

Atry: @cpcloud I think excluding `vllm-nccl-cu12` is the easiest approach. When `VLLM_NCCL_SO_PATH ` is unset, `vllm` will use `libnccl.so.2` from rpath as a fallback, according to https://github.com/vllm-project/vllm/blob/e8cc7967ff8a6f8432747a9e87ab451d36e1ff57/vllm/utils.py#L564C42-L564C54 . The default `libnccl.so.2` can be overridden easily in nix. We don't actually need `vllm-nccl-cu12`.

overrides/default.nix
Atry: `final.fastapi-cli` does not exist if `fastapi` version is 0.110.3

cpcloud: In that case, don't branch on a dependency existing, branch on the version of `fastapi`.

overrides/default.nix
cpcloud: Ok, SGTM. Then simply excluding the dependency would work IIUC.

overrides/default.nix
Atry: I think `final?fastapi-cli` is more robust than `prev.fastapi.version == "0.111.0"` because we don't know whether and when the upstream would fix the circular dependency.

overrides/default.nix
```diff
@@ -888,8 +888,13 @@ lib.composeManyExtensions [
...
+        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) (lib.optional (final?fastapi-cli) final.fastapi-cli);
```
cpcloud, line 893:
```suggestion
        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) (lib.optionals (final ? fastapi-cli) [ final.fastapi-cli ]);
```

overrides/default.nix
```diff
@@ -3991,6 +3999,13 @@ lib.composeManyExtensions [
         (old: { buildInputs = old.buildInputs or [ ] ++ [ pkgs.libxcrypt ]; });
 
       vllm = prev.vllm.overridePythonAttrs (old: {
+        # vllm-nccl-cu12 will try to download NCCL 2.18.1 from the internet to
+        # the ~/.config/vllm/nccl/cu12 directory, which is not allowed in Nix,
+        # so we do it ourselves
+        #
+        # See https://github.com/vllm-project/vllm/issues/4224
+        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) [final.vllm-nccl-cu12];
```
cpcloud, line 4007:
```suggestion
        propagatedBuildInputs = removePackagesByName (old.propagatedBuildInputs or [ ]) [ final.vllm-nccl-cu12 ];
```

cpcloud (approved):

Atry:
@cpcloud Do you want to merge this PR?

cpcloud (approved):

PR #1636, opened 2024-05-06
Update torch to 2.1.2

Atry:
`torch` 2.1.2 depends on `packaging` according to https://github.com/pytorch/pytorch/blob/a8e7c98cb95ff97bb30a728c6b2a1ce6bff946eb/torch/torch_version.py#L24


cpcloud:
Not 100% sure whether the fix I pushed for the darwin builds is safe: can we wheels built against a later macos SDK with an earlier SDK? For example, can we use a macos 10.13 wheel with  macos 10.9 runtime SDK?

cpcloud:
I suppose the safe thing to do is to avoid building/testing `ml-stack` on macos

cpcloud:
aarch64-darwin builds work locally, though we don't yet have a way to test those in CI.

cpcloud (approved):

PR #1651, opened 2024-05-11
fix matplotlib override

vale981:
This little patch allows skipping the patching of a source file in matplotlib if a wheel is used, and it is thus not available. For me this "works" but maybe there is some objection to doing it this way.

Can `preferWheels = true` be detected/accessed inside the override?


cpcloud:
Fixed up the PR to detect the wheel in the expression.

vale981:
thank you so much!

PR #1688, opened 2024-06-08
overrides: add support and hashes for ruff 0.4.5+

celestefox:
This adds support for ruff versions v0.4.5+, which add a builtin language server, which in turn uses a git dependency, which cannot be built as previous ruff versions were.
There may be a better way to do this still, but I think this works well enough. Might want to consider moving the ruff locks into a separate directory, especially if the git dependency stays and we keep needing more?
And to be clear, if we don't have the local `Cargo.lock`s, these ruff versions cannot be built without import-from-derivation, while older versions can be built without IFD as long as there's both hashes.

- **feat: allow ruff `CargoHash`es to be attrsets**
- **overrides: add hash and lock for ruff 0.4.5**
- **overrides: add hash and lock for ruff 0.4.6**
- **overrides: add hash and lock for ruff 0.4.7**
- **overrides: add hash and lock for ruff 0.4.8**


cpcloud (approved):

cpcloud:
Thanks!

PR #1696, opened 2024-06-11
Fix: url-specified source packages

### LINKED ISSUE #1695: Hash Mismatch when specifing a package via url
I'm specifying a (source, python only) package via {url= "...} in my pyproject.toml.

I get a hash mismatch somewhere inside poetry2nix.
```
       error: hash mismatch in file downloaded from 'https://files.pythonhosted.org/packages/4b/43/0d1b0135841e4f2b9a3208df24abef52dd368be166bb150369d2cd6c9bed/dppd-0.24.tar.gz':
         specified: sha256:061mqynvg15ramcy40szqakd3jzsl1jfyifqxdzr2f5mqpkn4xlh
         got:       sha256:1yb30ngnqsqzdzi65qmnwkx8a7smm63hqay2mw3ndgcjjhv54mqv
```

The hash in poetry.lock seems correct - at least it matches what I get when I download the file myself.

### Additional context
I have prepped a minimal example:
https://github.com/TyberiusPrime/poetry_url_bug_maybe

sha256:061mqynvg15ramcy40szqakd3jzsl1jfyifqxdzr2f5mqpkn4xlh
seems to be the base32 representation of "sha256:907662e7c5b538917febd845ef64a0facbd1a6c25f03e25955b984b7adc73518" inside the poetry.lock file.

I have no clue where 1yb30... comes from.

Edit: but hacking 'sha256:1yb30ngnqsqzdzi65qmnwkx8a7smm63hqay2mw3ndgcjjhv54mqv' into poetry.lock makes it build.

Edit2: 1yb30 appears to be the fetchzip hash for that tar.gz. Which I suppose is the unzipped recursive hash, at least that's what the wiki claims.

TyberiusPrime:

- [x] There's an _[automated test](tests/default.nix)_ for this change in use-url-src

This one line change fixes #1695, and introduces a test case. I verified that the test case fails (locally) with `nix build -L ".#githubActions.checks.x86_64-linux.use-url-src"` before the change, and succeeds afterwards.

Let's see if the rest of the CI likes it.


mk-poetry-dep.nix
```diff
@@ -214,7 +214,7 @@ pythonPackages.callPackage
                   sha256 = fileInfo.hash;
                 }
             else if isUrl then
-              builtins.fetchTarball
+              builtins.fetchurl
```
cpcloud, line 217: What are the major differences between these two functions? Why is there hash mismatch when using one but not the other?

TyberiusPrime: - fetchurl fetches an url and stores whatever bytes comes back into one file in the nix store. Hash is the hash of the bytes.
- fetchTarball fetches an archive (can be zip as well, at least), and unpacks it into a folder in the nix store. Hash is the 'recursive hash' of all the unpacked files.

Poetry only records the archive's hash, not the recursive one - that's pretty nix specific in how it 's done.

cpcloud: Ok, then it sounds like `fetchurl` is a closer match for poetry's behavior.

cpcloud (approved):

TyberiusPrime:
Thank you @cpcloud

Issue #1587, opened 2024-04-11
pyee no longer contains setup.py; installation fails for recent versions


bddap:
[This line](https://github.com/nix-community/poetry2nix/blob/3c92540611f42d3fb2d0d084a6c694cd6544b609/overrides/default.nix#L3580) expects setup.py to be present. Installation fails as a result.

### Additional context

`shell.nix` (run `nix-shell`):
```nix
{ pkgs ? import <nixpkgs> { } }:
let
  poetry2nix = pkgs.callPackage (pkgs.fetchFromGitHub {
    owner = "nix-community";
    repo = "poetry2nix";
    rev = "3c92540611f42d3fb2d0d084a6c694cd6544b609";
    sha256 = "2GOiFTkvs5MtVF65sC78KNVxQSmsxtk0WmV1wJ9V2ck=";
  }) { };
in poetry2nix.mkPoetryApplication { projectDir = ./.; }
```

`pyproject.toml`:
```toml
[tool.poetry]
name = "issue"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.11"
pyee = "^11.1.0"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

`poetry.lock`:
```toml
# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.

[[package]]
name = "pyee"
version = "11.1.0"
description = "A rough port of Node.js's EventEmitter to Python with a few tricks of its own"
optional = false
python-versions = ">=3.8"
files = [
    {file = "pyee-11.1.0-py3-none-any.whl", hash = "sha256:5d346a7d0f861a4b2e6c47960295bd895f816725b27d656181947346be98d7c1"},
    {file = "pyee-11.1.0.tar.gz", hash = "sha256:b53af98f6990c810edd9b56b87791021a8f54fd13db4edd1142438d44ba2263f"},
]

[package.dependencies]
typing-extensions = "*"

[package.extras]
dev = ["black", "build", "flake8", "flake8-black", "isort", "jupyter-console", "mkdocs", "mkdocs-include-markdown-plugin", "mkdocstrings[python]", "pytest", "pytest-asyncio", "pytest-trio", "sphinx", "toml", "tox", "trio", "trio", "trio-typing", "twine", "twisted", "validate-pyproject[all]"]

[[package]]
name = "typing-extensions"
[...]

PR #1710, opened 2024-06-20
overrides: pyee>10.0.0 fixed

TyberiusPrime:
Pyee removed their setup.py (in a setuptools project), and the old patch thefore fails

Closes: #1587


overrides/default.nix
```diff
@@ -2158,6 +2158,24 @@ lib.composeManyExtensions [
         ];
       });
 
+      polars = prev.polars.overridePythonAttrs (
+        old: {
+           cargoDeps = let
+             unpacked_src = pkgs.runCommand "polars-cargo-lock" {} ''
```
cpcloud, line 2164: Does this use IFD? I can't recall what the exact criteria for that are.

overrides/default.nix
```diff
@@ -3847,11 +3865,13 @@ lib.composeManyExtensions [
       # manually specify its version.
       pyee = prev.pyee.overrideAttrs (
         old: {
-          postPatch = old.postPatch or "" + ''
+          postPatch = old.postPatch or "" +
+          (if lib.versionOlder old.version "10.0.0" then
```
cpcloud, line 3869: You can use `lib.optionalString` here for a bit more concision.

TyberiusPrime:
I must have mixed up my git branches, the polar's stuff is not supposed to be in this pr. 
Sorry!.

I also think it's not IDF, since there's no nix code being imported. 

lib.optionalString is of course a good idea.

overrides/default.nix
TyberiusPrime: I can probably even skip the parenthesis, but I'm always wary about nix order of operations.

cpcloud (approved):

PR #1739, opened 2024-07-16
Add override for Pendulum version 3.0.0 with Rust

bow:
Hi maintainers,

I recently came across build problems when trying to use `poetry2nix` on a Python project that depends on [Pendulum](https://pendulum.eustace.io/). The build error was because this was the [first Pendulum release that has parts of its source re-written in Rust](https://github.com/sdispater/pendulum/releases/tag/3.0.0). Previous versions have been pure Python.

This PR fixes that issue by adding an override for Pendulum that includes the Rust build hook. The override is only applied if the version is at least 3.0.0 (still the most recent version at the time of writing).

The changeset of this PR works locally and I have also included a test for this, to confirm also that builds of previous versions are not broken. That said, I did encounter an issue where I found out the `version` attribute of [the autogenerated `Cargo.lock` in pendulum's source tree is actually `3.0.0-beta-1`](https://github.com/sdispater/pendulum/blob/3e3fec6a5b66759078c5ae39c736ca95b78adbef/rust/Cargo.lock#L7). This introduced some conflict when building.

Furthermore, I found out that [the commit on `master` on which the GitHub release was done](https://github.com/sdispater/pendulum/commit/3e3fec6a5b66759078c5ae39c736ca95b78adbef) is actually different from [the `3.0.0`-tagged commit](https://github.com/sdispater/pendulum/commit/0fcd10217af0469b3edda072f2b152d5273f3d58). One of the difference being, that the `[project].version` value in `pyproject.toml` of those two commits differ (`3.0.0` vs `3.0.0b1`). I suspect this was the cause of the differing `version` value in the `Cargo.lock` file.

~~These differences, as noted in the code, led me to the use of `fetchCargoTarball` instead of `importCargoLock`. The latter would require some manual string replacemend (e.g. in `prePatch` phase), which I think complicates the override too much.~~

EDIT: Apparently `fetchCargoTarball` still produces the same error (hash mismatch). My local builds were passing, but it seemed to be caused by some intermediate derivation not required for the final build. CI says otherwise, and when I run garbage collection, I bumped on the same error again. So now, I am using `importCargoLock` with an explicit `Cargo.lock` file added as part of this MR. Turns out this simplifies the build, in that we do not need to download and patch the source tree anymore.

Anyway, I hope this is useful ~ and feel free to critique, this is my first contribution to the repo :slightly_smiling_face:.


bow:
An additional question not directly related to this PR: During the course of playing with this package (that depends on pendulum), I found several other Python packages that needed `setuptools` as their build dependency.

I suspect the fix for these is just new entries in `overrides/build-systems.json`.

So for these packages, do you prefer a new PR altogether or is it ok to just add commits in this PR?

overrides/default.nix
```diff
@@ -2097,6 +2097,29 @@ lib.composeManyExtensions [
...
       );
 
+      pendulum = prev.pendulum.overridePythonAttrs (
+        old: (
+          # NOTE: Versions <3.0.0 is pure Python and is not PEP-517 compliant,
+          #       which means they can not be built using recent Poetry versions.
+          if lib.versionOlder old.version "3"
+          then { }
+          else
+            lib.optionalAttrs (!old.src.isWheel or false) {
```
cpcloud, line 2107: You don't need both the `if then else` **and** `lib.optionalAttrs`, just roll the `versionOlder` condition in to the first argument of `lib.optionalAttrs`. Feel free to parenthesize if that makes rolling it in easier.

cpcloud (changes requested):

overrides/default.nix
cpcloud: Fixed it up.

bow: Hey @cpcloud ~ thanks for the input (this was my first time going deeper into Nix). Is there anything that I should do from my end now, seeing that this change is already in the branch?

cpcloud (approved):

PR #1785, opened 2024-09-02
Treat default dependencies as main group

### LINKED ISSUE #157: ansible doesn't build (cannot find jinja2 - even though it appears to be present)
poetry2nix fails to build when the `ansible` dependency is included, even though poetry itself has no problems building that dependency.

```
[tool.poetry.dependencies]
python = "^3.6"
ansible = "~2.7"
```

(for full details on pyproject.toml and poetry.lock, see also https://github.com/nix-community/poetry2nix/pull/156/files) 

```
/nix/store/yglcb0pcpyh20139fv6q9si0zgxpzsv2-python3.8-molecule-2.22.drv
  /nix/store/wygkjnvbp9fxjp49syx3lm0vi7yfsidb-python3-3.8.3-env.drv
building '/nix/store/67qy42b85fbgqbkknz37ycsizrppgf8v-python3.8-ansible-2.7.18.drv'...
Sourcing python-recompile-bytecode-hook.sh
Sourcing pip-build-hook
Using pipBuildPhase
Using pipShellHook
Sourcing pip-install-hook
Using pipInstallPhase
Using pythonImportsCheckPhase
unpacking sources
unpacking source archive /nix/store/dq54rsvfnim1jjd87qb02g6mksda9v2p-ansible-2.7.18.tar.gz
source root is ansible-2.7.18
setting SOURCE_DATE_EPOCH to timestamp 1589250017 of file ansible-2.7.18/PKG-INFO
patching sources
configuring
no configure script, doing nothing
building
Executing pipBuildPhase
Creating a wheel...
Processing /build/ansible-2.7.18
Building wheels for collected packages: ansible
  Building wheel for ansible (setup.py) ... done
  Created wheel for ansible: filename=ansible-2.7.18-py3-none-any.whl size=9442989 sha256=c96f0c39e3091a1fb823eaeba75af32891e3abe88dcab7221074a8790a763f90
  Stored in directory: /homeless-shelter/.cache/pip/wheels/57/b5/be/8c108fa93269e4adfeed3fa84af37d6cfab90d2afdb8673f3f
Successfully built ansible
Finished creating a wheel...
Finished executing pipBuildPhase
installing
Executing pipInstallPhase
/build/ansible-2.7.18/dist /build/ansible-2.7.18
Processing ./ansible-2.7.18-py3-none-any.whl
ERROR: Could not find a version that satisfies the requirement jinja2 (from ansible==2.7.18) (from versions: none)
ERROR: No matching distribution found for jinja2 (from ansible==2.7.18)
builder for '/nix/store/67qy42b85fbgqbkknz37ycsizrppgf8v-python3.8-ansible-2.7.18.drv' failed with exit code 1
cannot build derivation '/nix/store/wygkjnvbp9fxjp49syx3lm0vi7yfsidb-python3-3.8.3-env.drv': 1 dependencies couldn't be built
error: build of '/nix/store/wygkjnvbp9fxjp49syx3lm0vi7yfsidb-python3-3.8.3-env.drv' failed
```

Including jinja2 as a dependency explicitly does not help. Same also happens when using latest ansible 2.9.x.

### LINKED ISSUE #1578: Default dependencies should work as if they were in `main` group
### Describe the issue

Poetry documentation on dependency groups [says](https://python-poetry.org/docs/managing-dependencies#dependency-groups):

> The dependencies declared in `tool.poetry.dependencies` are part of an implicit `main` group.

This is quite important when users interact with the project both via `poetry2nix` and via plain `poetry`:
1. when a non-Nix user runs `poetry add dependency`, `dependency` appears in default group
2. if we have a `dev` group, Nix user would want `{ groups = [ "main" "dev" ]; }` in their devShell
3. but `groups` work only for named groups, and default group is not treated as `main`
4. thus, we have to do `poetry add dependency --group main` (which non-Nix users will forget to do, and Nixers will have to fix it later for them :))

### Expected behaviour

It would be great if `groups` would either accept e.g. `null` as label for default dependency group, or treat `main` as default group.

P.S.: Thank you for making this tool!

Lykos153:
Poetry documentation on dependency groups [1] says:
  "The dependencies declared in tool.poetry.dependencies are part of an
implicit main group."

The current implementation does not account for that and in particular does not allow to skip installation of default dependencies like in one of
`poetry install --without main`
`poetry install --only-root`
`poetry install --only othergroup`.

This commit adds main as an implicit group and also adds it as a default value where applicable.

[1]
https://python-poetry.org/docs/managing-dependencies#dependency-groups

Closes #1578

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,7 +31,7 @@ let
     , pyProject
     , attrs
     , includeBuildSystem ? true
-    , groups ? [ ]
+    , groups ? [ "main" ]
     , checkGroups ? checkGroupsDefault
     , extras ? [ "*" ]  # * means all extras, otherwise include the dependencies for a given extra
     }:
@@ -82,12 +82,9 @@ let
     {
       buildInputs = mkInput "buildInputs" (if includeBuildSystem then buildSystemPkgs else [ ]);
       propagatedBuildInputs = mkInput "propagatedBuildInputs" (
-        getDeps allRawDeps ++ (
-          # >=poetry-1.2.0 dependency groups
-          if pyProject.tool.poetry.group or { } != { }
-          then lib.flatten (map (g: getDeps pyProject.tool.poetry.group.${g}.dependencies) groups)
-          else [ ]
-        )
+        let
+          availableGroups = {main.dependencies = allRawDeps;} // pyProject.tool.poetry.group or { };
+        in lib.flatten (map (g: getDeps availableGroups.${g}.dependencies) groups)
       );
       nativeBuildInputs = mkInput "nativeBuildInputs" [ ];
       checkInputs = mkInput "checkInputs" checkInputs';
@@ -145,7 +142,7 @@ lib.makeScope pkgs.newScope (self: {
       # Example: { my-app = ./src; }
     , editablePackageSources ? { }
     , pyProject ? readTOML pyproject
-    , groups ? [ ]
+    , groups ? [ "main" ]
     , checkGroups ? checkGroupsDefault
     , extras ? [ "*" ]
     }:
@@ -314,7 +311,7 @@ lib.makeScope pkgs.newScope (self: {
     , preferWheels ? false
     , editablePackageSources ? { }
     , extraPackages ? _ps: [ ]
-    , groups ? [ "dev" ]
+    , groups ? [ "main" "dev" ]
     , checkGroups ? checkGroupsDefault
     , extras ? [ "*" ]
     }:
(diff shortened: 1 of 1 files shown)


Lykos153:
A more backwards compatible solution would be to add a flag like `useDefaultDeps ? true`. Though in this case the behavior is not completely obvious, because poetry also installs by default all groups that don't have `optional = true`.

Lykos153:
Marking as ready for review in order to get feedback on whether the approach in this PR would be accepted or if a flag as proposed in my last comment would be preferred.

cpcloud:
The initial approach here seems reasonable to me.

cpcloud:
Thanks!

PR #1796, opened 2024-09-06
Upstream overrides

PerchunPak:
These are overrides I made for my projects, just upstreaming them


overrides/default.nix
```diff
@@ -1387,6 +1388,66 @@ lib.composeManyExtensions [
...
+              owner = "Instagram";
+              repo = "LibCST";
+              rev = old.version;
+              sha256 = githubHash;
+            };
+
+          in
+          {
+            inherit src;
+            cargoDeps = pkgs.rustPlatform.importCargoLock {
```
adisbladis, line 1438: This is using IFD. Please use `cargoVendorHash` instead.

PerchunPak: You mean `cargoHash`? I cannot find anything about `cargoVendorHash` in nixpkgs

PerchunPak: I assume you meant `pkgs.rustPlatform.fetchCargoTarball`, but why? `importCargoLock` gives an ability to have less maintenance burden because of not having hash for every single release. This would still be reproducible, as hashes are in poetry.lock, but also will make that any new version of libcst is automatically supported:

```nix
libcst = prev.libcst.overridePythonAttrs (
  old: lib.optionalAttrs (!(old.src.isWheel or false)) (
    let
      unpackedSrc = stdenv.mkDerivation {
        name = "unpacked-${old.src.name}";
        inherit (old) src;
        buildPhase = "cp -r . $out";
      };
    in
    {
      cargoDeps = pkgs.rustPlatform.importCargoLock {
        lockFile = "${unpackedSrc.out}/native/Cargo.lock";
      };
      cargoRoot = "native";
      nativeBuildInputs = old.nativeBuildInputs or [ ] ++ [
        pkgs.rustPlatform.cargoSetupHook # handles `importCargoLock`
        pkgs.rustc
        pkgs.cargo
      ];
    }
  )
);
```

There are also multiple other places, where poetry2nix could automatically get dependencies, but instead you chose to manually add them to every package. E.g. if poetry or setuptools are required for build: https://github.com/py-mine/mcstatus/blob/0bbd4842178fa1db252371ce5599651e4b867efd/pyproject.toml#L164

spinus: @adisbladis can you elaborate?

PerchunPak: I learned about IFD since then and will try to fix the issue

PerchunPak:
Should I add an IFD alternative for versions older than 1.0.0? (libcst is not that popular after all)

```
$ du -sh overrides/libcst
348K	overrides/libcst
```

overrides/default.nix
```diff
@@ -1023,11 +1023,12 @@ lib.composeManyExtensions [
...
 
-      gitpython = prev.gitpython.overridePythonAttrs (
-        old: {
-          buildInputs = old.buildInputs or [ ] ++ [ final.typing-extensions ];
-        }
-      );
+      gitpython = prev.gitpython.overridePythonAttrs {
+          postPatch = ''
+            substituteInPlace git/cmd.py \
+              --replace 'git_exec_name = "git"' 'git_exec_name = "${pkgs.gitMinimal}/bin/git"'
```
elikoga, line 1029: gitpython should not decide on it's own to pull in a specific git as a dependency imo

PerchunPak: How it will function then? It is a general approach to patch packages to provide their dependencies without propagating them. You can change git version used using an overlay

elikoga: Provide the git binary in `PATH` or override the variable in your application code

elikoga: I do not expect `gitpython` the `python` package to install `git` for me

PerchunPak: Well, this is how gitpython in nixpkgs is packaged:

https://github.com/NixOS/nixpkgs/blob/a8be94c0af330325c8c96730cdae5e168a0632a1/pkgs/development/python-modules/gitpython/default.nix#L34

elikoga: Looks like it was introduced 6 years ago in upstream: https://github.com/NixOS/nixpkgs/pull/41844

I guess I'm fine with this behaviour then

PR #1799, opened 2024-09-08
fix(shellcheck-py): Support 0.10

l0b0:
WIP failing test because of setuptools-download being run at build time.


tests/default.nix
```diff
@@ -173,6 +173,7 @@ in
   soundfile-wheel = callTest ./soundfile-wheel { };
   soundfile = callTest ./soundfile { };
   pyogrio = callTest ./pyogrio { };
+  shellcheck-py = callTest ./shellcheck-py { };
```
cpcloud, line 176: This package seems a bit pointless if you're using nix.

It doesn't make shellcheck importable, it's just a package to download and install the right shellcheck binary **using** Python.

If you're using nix, can't you just use `pkgs.shellcheck` and call it a day? 😂

l0b0: If you're *only* using Nix, yes. But if you work with people not using Nix, you might have `shellcheck-py` in `poetry.lock`, and then it's annoying having to figure out how to replace that with `pkgs.shellcheck` in `shell.nix` and the like.

cpcloud (approved):

PR #1828, opened 2024-09-27
[fix] change path to defaultPoetryOverrides in the doc

edelagnier:

Having tried to override a buildInputs according to the edgecase documentation, I was confronted to this error : 

```
       error: attribute 'defaultPoetryOverrides' missing
```

I then found [this question](https://discourse.nixos.org/t/need-help-on-moving-python-poetry-project-to-flake-nix/41356) to which someone suggested to use "poetry2nix.lib.mkPoetry2Nix.defaultPoetryOverrides instead of poetry2nix.defaultPoetryOverrides" 

It did work for me, so I'm creating this PR but is this the way it should be handled ? or would you suggest another solution ?

### DIFF
--- a/docs/edgecases.md
+++ b/docs/edgecases.md
@@ -127,7 +127,7 @@ In order to be able to build `django-floppyforms` we should modify our nix defin
 ``` nix
 poetry2nix.mkPoetryApplication {
   projectDir = ./.;
-  overrides = poetry2nix.defaultPoetryOverrides.extend
+  overrides = poetry2nix.lib.mkPoetry2Nix.defaultPoetryOverrides.extend
     (final: prev: {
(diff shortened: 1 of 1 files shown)


cpcloud:
Do you have some code that reproduces the problem? I use `poetry2nix.defaultPoetryOverrides` just fine [here](https://github.com/ibis-project/ibis/blob/main/nix/ibis.nix#L24), as well as [here](https://github.com/ibis-project/ibis/blob/main/nix/overlay.nix#L9).

Is it possible you're missing some setup code like [this](https://github.com/ibis-project/ibis/blob/main/flake.nix#L14-L28)?

Nebucatnetzer:
As @cpcloud pointed out the README is already correct.
Feel free to open an issue if you still experience an issue with this.

PR #1836, opened 2024-10-02
Several override additions and fixes

cheriimoya:


overrides/default.nix
```diff
@@ -4279,9 +4309,11 @@ lib.composeManyExtensions [
       });
 
       mkdocs-material = prev.mkdocs-material.overridePythonAttrs (old: {
-        postPatch = old.postPatch or "" + ''
-          sed -i 's/"Framework :: MkDocs",//' pyproject.toml
+        postPatch = old.postPatch or "" + lib.optionalString (old.version == "8.5.4") ''
+          sed -i 's/filename = "requirements.txt"//' pyproject.toml
+          sed -i '/\[project\]/a dependencies = ["jinja2>=3.0.2", "markdown>=3.2", "mkdocs>=1.3", "mkdocs-material-extensions>=1.0.3", "pygments>=2.12", "pymdown-extensions>=9.4", "requests>=2.26"]' pyproject.toml
```
cheriimoya, line 4314: Not sure, why the `"Framework :: MkDocs",` was `sed`ed away, it works without for me. This solution doesn't look pretty, but I don't know how else to solve the problem (see commit message).

cpcloud: Do you have a `pyproject.toml` that fails without the second sed line?

A simple test derivation that only depends on `mkdocs-material = "8.5.4"` builds for me with **only** the line that removes `filename = "requirements.txt"`.

cpcloud: I pushed up a simpler replacement that only replaces `filename = "requirements.txt"`. Let me know if that works for you, and then we can merge this PR.

cheriimoya: I wanted to double check, if everything works without specifying the dependencies, like coloring of code with pygments, but `poetry lock` already resolves the dependencies. So I guess it's fine. Yet, I added mkdocs-get-deps and now something on darwin fails. Cannot reproduce locally as I don't have darwin. Can also be merged without the last commit.

PR #1884, opened 2025-01-07
Remove Leaked `indirect` Input

benj2468:
Removes the leaked indirect `inputs.systems`. This created an ambiguous and indirect lockfile entry, this is resolved with this PR, which creates a clear `original` entry with fields similar to the `locked` entry. 

Furthermore, not having it marked as an input is not inline with the examples of the `nix-systems` project found [here](https://github.com/nix-systems/nix-systems/blob/main/examples/simple/flake.nix#L4). 

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -88,8 +88,9 @@
         "type": "github"
       },
       "original": {
-        "id": "systems",
-        "type": "indirect"
+        "owner": "nix-systems",
+        "repo": "default",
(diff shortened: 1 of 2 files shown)


elikoga (approved):
Good catch!

This means that `systems` will not get resolved via https://github.com/NixOS/flake-registry but directly. And since we actually depend on it, a explicit reference in the inputs is good.

Issue #1174, opened 2023-06-04
Poetry2nix still resolves all dependencies with groups/markers


leolavaur:
### Describe the issue

I am trying to build a cross-platform flake template for Python and TensorFlow using Poetry2nix. Issues arise especially on MacOS and Apple Silicon architecture, as using TensorFlow requires different packages (`tensorflow-macos` and `tensorflow-metal`, instead of `tensorflow`). I am writing a "generic" `pyproject.toml` file for that purpose, but since `tensorflow` does not have any compatible release with `aarch64-darwin`, the execution will fail with 
```
error: Missing suitable source/wheel file entry for tensorflow
```

I first tried to use markers (with `"sys_platform == 'darwin'"`), but it failed. I decided to move the logic of platform selection to my `flake.nix`, and use Poetry groups to install the correct dependencies when using `mkPoetryApplication`/`mkPoetryEnv`, but it still yields the same error.

From what I can understand, neither groups nor markers appear in the lock file. Poetry2nix first parses the lockfile and gathers dependencies, and then figures out which to install. Is there a solution for what I'm trying to do?

Thanks!

### Additional context

You can find the repository here: https://github.com/phdcybersec/python-nix/tree/tensorflow-apple-silicon.
Bellow are the details.

<details><summary>Code samples</summary>
<p>
`flake.nix`:

```nix
{
  # inspired by: https://medium.com/@sorenlind/tensorflow-with-gpu-support-on-apple-silicon-mac-with-homebrew-and-without-conda-miniforge-915b2f15425b
  # ---
  # 1. look at https://anaconda.org/apple/tensorflow-deps for the list of dependencies in the
  #   meta.yaml file
  # 2. add hdf5 to build inputs
  # 3. install the python dependencies with poetry, eg. for tensorflow-deps 2.10:
  #    - grpcio >=1.37.0,<2.0
  #    - h5py >=3.6.0,<3.7
  #    - numpy >=1.23.2,<1.23.3
  #    - protobuf >=3.19.1,<3.20

  description = "Tensorflow devShell for MacOS with poetry2nix";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-22.11";

    poetry2nix = {
      url = "github:nix-community/poetry2nix";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = { self, nixpkgs, poetry2nix, ... }:
    let pythonVer = "python310"; in
    {
[...]

PR #1905, opened 2025-03-11
fix: dependencies package PEP508 markers selection

jamesdbrock:
Bugfix: support local `.whl` files with dependency markers.

Maybe resolves https://github.com/nix-community/poetry2nix/issues/1174

2 commits:

1. Add 2 tests.  Both tests are disabled for the case of `x86_64-darwin`.
   -  **Remote `.whl` dependency marker selection** `mk-poetry-packages-partition-whl-remote` On this commit the test succeeds because *poetry2nix* can select dependency markers for remote `.whl` files.
   - **Local `.whl` dependency marker selection** `mk-poetry-packages-partition-whl-local` On this commit the test fails because *poetry2nix* cannot select dependency markers for local `.whl` files.

     Add three local `watchdog.whl` files (79kB, 79kB, 96kB) to the repository. We chose *watchdog* because it distributes separate `.whl` files for each platform and the files are small.

2. fix: dependencies package PEP508 markers selection.
   With this commit the second test  `mk-poetry-packages-partition-whl-local` succeeds because now *poetry2nix* can select dependency markers for local `.whl` files.

[Contribution](https://github.com/nix-community/poetry2nix/blob/master/README.md#contributing) checklist (recommended but not always applicable/required):

- [x] There's an _[automated test](https://github.com/nix-community/poetry2nix/blob/master/tests/default.nix)_ for this change


jamesdbrock:
Who is merging PRs on this repo these days, is it @Nebucatnetzer ? Thank you very much for your maintenance work.

This PR might be controversial because it adds 3 `.whl` binary files to the repo. I'm willing to negotiate about that.

Nebucatnetzer:
I can merge stuff but that mostly involves the pull requests for override updates and similar minor stuff.

I honestly don't know the pros and cons of including the wheels here.
Maybe they could be included with something like fetchurl?

When you write `Maybe resolves #1174` what makes you unsure about?
I guess you tested it?

Nebucatnetzer:
Maybe @elikoga has some inputs?

elikoga:
I was initially a bit confused because the diff changes a read for `.marker` to `.markers` but it seems like the poetry lock format uses markers, as shown in the patch.

I'd look into why we accessed `.marker` before or wether that was just buggy code.

I'm not sure about adding the wheels to the repo. They are small (all less than 100kb as far as I see) so it's not that bad.

jamesdbrock:
> I'd look into why we accessed `.marker` before or wether that was just buggy code.

Yeah that's a good question and I don't have a good answer. It looks like we've been [using `.marker` since it was introduced in 2019](https://github.com/nix-community/poetry2nix/commit/ff8d58af7dbc286923139d664a32f986a8fe9756#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R96).

My guess is that it was correct at that time, but for some reason the `pkgMeta`, which is read from the `poetry.lock` TOML, has changed.

https://github.com/nix-community/poetry2nix/blob/f842aae3a72ef764e38eaf3e75aa435fb9861bba/default.nix#L166

But there is something wrong with my guess because [Poetry only started including `markers` in the `poetry.lock` two months ago in v2.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/#locking-resulting-markers-and-groups).

jamesdbrock:
> I honestly don't know the pros and cons of including the wheels here. Maybe they could be included with something like fetchurl?

I don't think the `.whl` files can be included by fetchurl because if we fetchurl them then they are placed in the Nix Store. But there is no way to specify the Nix Store path in the [`pyproject.toml`](https://github.com/nix-community/poetry2nix/blob/e5ebc9402cf42121c70f4e1bf6dddb9b18dccf97/tests/mk-poetry-packages-partition-whl-local/pyproject.toml).

jamesdbrock:
> When you write `Maybe resolves #1174` what makes you unsure about? I guess you tested it?

Because #1174 was opened by someone else and their problem is actually slightly different than my problem. The tests included in this PR prove that this PR solves my problem. I think this PR might also solve their problem but I'm not sure.

jamesdbrock:
Another question:

Why is it currently true that *poetry2nix* can select dependency markers for remote `.whl` files, but *poetry2nix* cannot select dependency markers for local `.whl` files?

I don't know.

jamesdbrock:
I'm going to dynamically generate the `pyproject.toml`; putting this back in Draft mode.

jamesdbrock:
By the way the current gzipped size of *poetry2nix* is 2.1MB.

image

and the total uncompressed size is 8.6MB.

image

This PR will add about 300kB because of the `.whl` files.

jamesdbrock:
Another option is that we could get rid of the tests and just merge the patch commit.

Nebucatnetzer:
> I'm going to dynamically generate the `pyproject.toml`; putting this back in Draft mode.
> 
> EDIT
> 
> No, come to think of it, to use `fetchurl` we would have to dynamically generate both the `pyproject.toml` and the `poetry.lock` with the Nix Store paths of the fetched `.whl` files. Dynamically generating both of those files is not worth the complication, in my opinion.

I agree, lets not make it too complicated.

Nebucatnetzer:
> Another option is that we could get rid of the tests and just merge the fix commit. That would not increase the size of the repo. And do we really need those tests?

Well it looks like we didn't notice the change from `marker` to `markers` now we would have a test that uses them so I think that is an improvement.

PR #1925, opened 2025-11-27
chore: update vendored pyproject.nix

poelzi:
Updates the vendored pyproject.nix to the latest version, which includes:

- riscv64 support in manyLinuxTargetMachines (pep599.nix)
- Simplified parseABITag that handles all ABI tags including those with underscores like 'graalpy242_311_native' (pypa.nix)
- Various documentation and code improvements

This fixes wheel compatibility checking for RISC-V builds and GraalPy interpreters.

### DIFF
--- a/vendor/pyproject.nix/default.nix
+++ b/vendor/pyproject.nix/default.nix
@@ -5,4 +5,9 @@ lib.fix (self: {
     pyproject-nix = self;
     inherit lib;
   };
+  packages =
+    { callPackage }:
+    {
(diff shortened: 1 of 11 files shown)


sdaqo:
Tests are failing, should be `lib.nix` instead of `util.nix` in `lib.nix` I think:
```nix
{ lib, pyproject-nix, pkgs, ... }:
let
  inherit (import ./vendor/pyproject.nix/lib/util.nix { inherit lib; }) splitComma;

  fromTOML = builtins.fromTOML or

```

